PR #30, opened 2017-09-26
Fix UB in pin failure timeout printing

jroelofs:
The Undefined Behavior happens when failed_cnts > 31. See C99/C11: 6.5.7p3.

Even though it'll take ~272 years to reach failed_cnts == 32, the compiler is
allowed to assume this cannot happen, and optimize to that effect leading to
Bad Things (TM). Better to avoid that entirely.

Also, added benefit: the time displayed is now correct once failed_cnts == 32, should anyone that patient live long enough to visit the year 2244.

### DIFF
--- a/keepkey/local/baremetal/pin_sm.c
+++ b/keepkey/local/baremetal/pin_sm.c
@@ -298,12 +298,12 @@ bool pin_protect(char *prompt)
         {
             if(failed_cnts > 2)
             {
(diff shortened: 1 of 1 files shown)


keepkeyjon:
https://github.com/keepkey/keepkey-firmware/commit/4cf1c80fb2a94d5f91b60285cf28fb12fd9518a8

PR #39, opened 2018-01-31
Initial Erc20 Support

bgok:
ERC20 support for top tokens 
Position sensitive data to a position in the RAM where they will be overwritten on a reboot.
Wipe the RAM on startup


interface/public/messages.options
```diff
@@ -99,10 +99,14 @@ EthereumSignTx.address_n		max_count:8
 EthereumSignTx.nonce			max_size:32
 EthereumSignTx.gas_price		max_size:32
 EthereumSignTx.gas_limit		max_size:32
-EthereumSignTx.to			max_size:20
+EthereumSignTx.to			    max_size:20
 EthereumSignTx.value			max_size:32
 EthereumSignTx.data_initial_chunk	max_size:1024
 EthereumSignTx.to_address_n		max_count:8
+EthereumSignTx.chain_id         max_size:32
```
solipsis: We should get a c formatter we all use before the next release

keepkey/local/baremetal/util.c
```diff
@@ -103,3 +103,36 @@ void dec64_to_str(uint64_t dec64_val, char *str)
     sbfr++;
 }
 
+
+/*convert hex to bytes*/
+int hex0xstr_to_char(const char *hex_str, unsigned char *byte_array, int byte_array_max)
```
solipsis: This method is not used anywhere. Do we still want to include it?

keepkey/public/util.h
```diff
@@ -38,4 +39,7 @@ uint32_t readprotobufint(uint8_t **ptr);
 void rev_byte_order(uint8_t *bfr, size_t len);
 void dec64_to_str(uint64_t dec64_val, char *str);
 
+int hex0xstr_to_char(const char *hex_str, unsigned char *byte_array, int byte_array_max);
```
solipsis: keep or remove depending on what we do to the implementation above

interface/public/messages.options
bgok: Added retrospective topics on the sprint board

Issue #17, opened 2017-07-07
How to prove signed firmware from KeepKey?


Aragonite:
Trevor has a process that anyone can follow to prove that their signed firmware binary was made of the github source code at a specific tag.   I don't see the tools available with KeepKey to prove the same.

For example, with Trezor, using the script firmware-fingerprint.sh, will remove the first 256 header bits from their official signed binary (where their signature is located) and a SHA256SUM of the rest of the file should match the SHA256SUM of the binary that can be compiled independently with the Github source code (using their script firmware-docker-build.sh).

The problem I am having with KeepKey, is their docker_build_release.sh script produces a binary file that has no header (the first 256 bits with the magic code KPKY is missing) - therefore, that binary file can't even be loaded into a KeepKey unit.  

Question 1:  How can we "sign" a firmware that we have compiled ourselves and load it to a KeepKey unit - with the understanding that unit will warn us every time we turn on the device that the firmware is not signed by KeepKey.

Question 2:  How can we verify that the binaries provided by KeepKey matches the code on Github, for each version?

yate:
For question 2, I believe it's a similar process to trezor. 

- Download the signed binary from a keepkey release and run `tail -c +257 keepkey_main.bin | sha256sum`
- Download the source code for the release version and run `./docker_build_release.sh`. The script outputs the sha256sum of the built firmware which should match with the above.

I'm interesting in the question 1 as well. I would like to have a developer guide on how to push your built firmware to your device and how to debug your code.

keepkeyjon:
For question 1:

You'll need to arrange for this metadata to be prepended to your binary (in order):

  * 4 bytes of magic: `"KPKY"`.
  * 4 bytes of firmware payload length, encoded as a little-endian uint.
  * 3 bytes of signature indexes. I suggest just using: `0x01, 0x02, 0x03`.
  * 1 byte of flags: `0x01`
  * 52 reserved bytes, all `0x00`
  * 3x bogus signatures, 64 bytes each. I believe this can either be all `0x00`s, or any random garbage. Your pick.
  * Followed by the firmware payload itself (i.e. `keepkey_main.bin`).

After "signing", you can upload with:

```
$ keepkeyctl wipe_device
$ keepkeyctl firmware_update --file path/to/bin/debug/keepkey/keepkey_main.bin
```

[...]

PR #48, opened 2018-04-23
BTG Support

StarbuckBG:
Hello KeepKey team!
I have added configuration for BTG. I am not able to test it because of the firmware signing that you do. 

- Mind that I included the BTG forkId (79), but I'm not sure if it is on the right place in the config.

Also mind that currently the device can be tested with BTG version of Electrum (called ElectrumG), you can check it from here - https://github.com/BTCGPU/electrum .

### DIFF
--- a/keepkey/local/baremetal/coins.c
+++ b/keepkey/local/baremetal/coins.c
@@ -28,10 +28,13 @@
 
 /* === Variables =========================================================== */
 
+
+//Coin name, coin shortcut, address type p2pkh, max fee, address type p2sh, address type p2wpkh, address type p2wsh, signed message header, bip44_account_path, forkId
 const CoinType coins[COINS_COUNT] = {
         {true, "Bitcoin",     true, "BTC",  true,   0, true,     100000, true,   5, true,  6, true, 10, true,  "\x18" "Bitcoin Signed Message:\n",  true, 0x80000000, false, 0, true,   8, false, {0, {0}},                                          false, {0, {0}}},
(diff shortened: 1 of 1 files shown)


keepkeyjon:
@StarbuckBG / @wilsonmeier Thanks for the patch! BTG support was added here: https://github.com/keepkey/keepkey-firmware/commit/87c961bd7dca5b05e6e5a45b11ee9d3bd0658cbd https://github.com/keepkey/keepkey-firmware/commit/4011ad0626d89254d1f301c65c3c86a61d1eb427 and has been released in our beta client: https://chrome.google.com/webstore/detail/keepkey-client-rc/ndpjhnboalogfiafbknodggihfiedibo

> I am not able to test it because of the firmware signing that you do.

Devices that have not had their bootloaders patched / updated are still able to run unsigned firmware. You'll need to prepend the `0x100` bytes of metadata (magic, code length, fake signatures, etc) described here: https://github.com/keepkey/keepkey-firmware/issues/17#issuecomment-370561098, and then upload using the python client.

StarbuckBG:
Hello @keepkeyjon !

Thanks for the hardwork! 

We have already checked and tested it. Also we have added support for KeepKey in our ElectrumG wallet (BTG version of Electrum).
We want to ask if it is possible to add testnet config for the next release? We need it in order to test safely KeepKey on our network and out network on KeepKey.

keepkeyjon:
We're a bit resource-constrained at the moment since the Features object has a limited maximum size, and the ever-growing coin table is approaching that limit. After that has been addressed, it should not be a problem to add a BTG testnet entry.

Mind sending a PR for that, and I'll merge it once we've excised the coin table from the Features object?

PR #80, opened 2019-01-28
Make UTXO signing more permissive w.r.t. segwit change

keepkeyjon:
This allows treating BIP44+BIP49+BIP84 paths with the same account
number & coin type as one big account.


keepkeyjon:
https://github.com/mycelium-com/wallet-android/issues/483

keepkeyjon:
followup here: https://github.com/keepkey/keepkey-firmware/pull/81

PR #81, opened 2019-01-29
Mixed mode segwit

keepkeyjon:


keepkeyjon:
This time it comes with a test: https://github.com/keepkey/python-keepkey/pull/46

PR #82, opened 2019-02-08
Add Nano support

roosmaa:
This PR adds Nano support to the KeepKey firmware.

~Regarding the exchange support. What would be the best way to test this? Any mock ExchangeType content that could be used?~

This pull request depends on:
- keepkey/trezor-crypto#1
- keepkey/device-protocol#21
- keepkey/python-keepkey#47


lib/firmware/fsm_msg_nano.h
```diff
@@ -0,0 +1,383 @@
...
+                layoutHome();
+                return;
+            }
+        }
+    }
+
+    // Sign hash and return the signature
+    uint8_t signature[64];
+    memset(signature, 0, sizeof(signature));
+    hdnode_sign_digest(node, block_hash, signature, NULL, NULL);
```
keepkeyjon: `fsm_getDerivedNode()` has the node as a static inside it, which would get overwritten by the call here: https://github.com/keepkey/keepkey-firmware/pull/82/files#diff-a51ea15f50216fc3c1b6614a895b8027R223

lib/firmware/fsm_msg_nano.h
```diff
@@ -0,0 +1,383 @@
...
+    blake2b_Update(&ctx, account_pk, sizeof(ed25519_public_key));
+    blake2b_Update(&ctx, parent_hash, 32);
+    blake2b_Update(&ctx, representative_pk, sizeof(ed25519_public_key));
+    blake2b_Update(&ctx, balance, 16);
+    blake2b_Update(&ctx, link, 32);
+
+    blake2b_Final(&ctx, out_hash, 32);
+}
+
+void nano_truncate_address(const CoinType *coin, char *str) {
```
keepkeyjon: I'd move these helpers to a new file: `lib/firmware/nano.c`, which makes them unit-testable.

lib/firmware/fsm_msg_nano.h
roosmaa: Nice catch! :+1:

keepkeyjon: Usual fix is to copy the node and save it for later. memzero under all the failure condition branches.

roosmaa: Since there seems to be some caching involved in that function, I was thinking of just invalidating the pointer and then calling the `fsm_getDerivedNode` before signing if `node == NULL`.

keepkeyjon: Or just wait to calculate it until it's needed.

keepkeyjon:
> * Implement exchange support?

We'll take care of that if/when we end up adding it to ShapeShift.

lib/firmware/fsm_msg_nano.h
roosmaa: Done and added some unit-tests as well.

roosmaa:
Rebased on v6.0.2 tag. Looks like I need to rebase once more on latest master :smile:

roosmaa:
Rebased on v6.0.2 tag.

roosmaa:
Rebased on master.

keepkeyjon:
I forgot to mention... besides the unittests, we usually write end-to-end integration tests in python-keepkey, which helps us avoid regressions in the future.

roosmaa:
Updated the `deps/python-keepkey` to point to the branch `feat/nano-support`.

lib/firmware/fsm_msg_nano.h
```diff
@@ -0,0 +1,309 @@
...
+
+    // Determine what type of prompt to display
+    bool needs_confirm = true;
+    bool is_transfer = false;
+    if (strlen(representative_address) > 0) {
+        // always confirm representative change
+    } else if (!is_send) {
+        // don't bother confirming pure receives
+        needs_confirm = false;
+    } else if (msg->has_address_type &&
```
keepkeyjon: I think this whole `else if` for transfers ought to just be `} else if (msg->link_recipient_n_count) {`, since the path gets checked in the *_to_string helpers.

keepkeyjon: I always felt like the way that's handled for UTXO coins was a bit odd, so no need to replicate that strange behavior here.

keepkeyjon: For exchange, if/when we add that, it'd be a call to `run_policy_compile_output` here, and set `needs_confirm = false;` if it succeeds.

lib/firmware/fsm_msg_nano.h
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    // Calculate current hash
+    if (msg->has_link_hash) {
+        memcpy(link, msg->link_hash.bytes, sizeof(link));
+    } else if (msg->link_recipient_n_count > 0) {
+        node = fsm_getDerivedNode(coin->curve_name, msg->link_recipient_n, msg->link_recipient_n_count, NULL);
+        if (!node) return;
+        hdnode_fill_public_key(node);
+        memcpy(link, &node->public_key[1], sizeof(link));
```
keepkeyjon: Should add a `memzero(node, sizeof(*node));` here to clear the secret from that static. If the block were invalid, we'd have left the secrets in it.

lib/firmware/fsm_msg_nano.h
```diff
@@ -0,0 +1,309 @@
...
+
+    // Sign hash and return the signature
+    if (!node) {
+        node = fsm_getDerivedNode(coin->curve_name, msg->address_n, msg->address_n_count, NULL);
+        if (!node) return;
+    }
+
+    uint8_t signature[64];
+    memset(signature, 0, sizeof(signature));
+    hdnode_sign_digest(node, block_hash, signature, NULL, NULL);
```
keepkeyjon: I think this should sign directly into the response object, avoiding the need for this stack temporary.

lib/firmware/fsm_msg_nano.h
roosmaa: Yeah, was trying to mimic the existing implementations, totally agree that address_type is kind of redundant here. This will also means I can get rid of the NanoSignTx.address_type which I didn't really like in the protocol def.

lib/firmware/fsm_msg_nano.h
roosmaa: Should the `node` content be zeroed at the end of the function as well then? or is that handled somewhere else?

keepkeyjon: Yeah, it should be zeroed out when no longer needed.

lib/firmware/fsm_msg_nano.h
roosmaa: Addressed in 1be9bf6

lib/firmware/fsm_msg_nano.h
roosmaa: Addressed in fe8cfee

lib/firmware/fsm_msg_nano.h
roosmaa: Changed with 2bcfe03

lib/firmware/app_layout.c
```diff
@@ -496,6 +496,50 @@ void layout_ethereum_address_notification(const char *desc, const char *address,
     layout_notification_icon(type, &sp);
 }
 
+/*
+ * layout_nano_address_notification() - Display nano address notification
+ *
+ * INPUT
+ *     - desc: description of address being shown (normal or multisig)
```
keepkeyjon: Does Nano have multisig?

lib/firmware/nano.c
```diff
@@ -0,0 +1,86 @@
...
+        return false;
+    }
+
+    // m/44' : BIP44
+    // m / purpose' / bip44_account_path' / account'
+    if (address_n[0] == (0x80000000 + 44)) {
+        mismatch |= address_n_count < 3;
+        mismatch |= address_n_count > 1 && (address_n[1] != coin->bip44_account_path);
+        mismatch |= address_n_count > 2 && (address_n[2] & 0x80000000) == 0;
+        mismatch |= address_n_count > 3 && (address_n[3] & 0x80000000) == 0;
```
keepkeyjon: What are `address_n[3]` and `address_n[4]` for in Nano derivation paths?

lib/firmware/nano.c
```diff
@@ -0,0 +1,86 @@
...
+    blake2b_Update(&ctx, account_pk, 32);
+    blake2b_Update(&ctx, parent_hash, 32);
+    blake2b_Update(&ctx, representative_pk, 32);
+    blake2b_Update(&ctx, balance, 16);
+    blake2b_Update(&ctx, link, 32);
+
+    blake2b_Final(&ctx, out_hash, 32);
+}
+
+void nano_truncate_address(const CoinType *coin, char *str) {
```
keepkeyjon: Is this truncation part of the protocol, or something done to make the addresses fit on the screen?

keepkeyjon: If the latter, then this ought to be done right before display in the two places where it's called now.

lib/firmware/fsm_msg_nano.h
```diff
@@ -0,0 +1,318 @@
...
+        needs_confirm = false;
+    } else if (msg->link_recipient_n_count > 0 &&
+               !nano_path_mismatched(coin, msg->link_recipient_n,
+                                     msg->link_recipient_n_count)) {
+        is_transfer = true;
+    }
+
+    if (needs_confirm) {
+        if (strlen(representative_address) > 0) {
+            if (!confirm(ButtonRequestType_ButtonRequest_ConfirmOutput,
```
keepkeyjon: Are representatives' addresses static enough to have a lookup table of the top few well-known ones? That'd allow us to put human-readable names on some of them, which I believe users would be more likely to correctly validate on-screen.

lib/firmware/app_layout.c
roosmaa: No it doesn't. Didn't notice that part when I was adapting the comment block for the Nano method. Will update that block.

lib/firmware/nano.c
roosmaa: It's not part of the protocol per se, but it is a convention in most of the Nano wallets out there. The 5 chars of the public key and 5 chars of the checksum provide enough information for the user to quickly verify that they entered the correct address for their representative/recipient.

lib/firmware/fsm_msg_nano.h
roosmaa: I guess the more popular ones are static enough.. but then it would have to match that displayed on the actual desktop wallet, and not sure how many wallets actually display a human readable representative address. So if we did include the lookup table, we'd have to display both the human readable label and the actual address.

keepkeyjon: Good point. To get around that, we could have a separate `NanoLookupRepresentative` / `NanoRepresentative` pair of messages that clients could use to probe the firmware for known representatives before signing. Let's hold off on that though, I don't want to make this more complicated than it needs to be.

lib/firmware/nano.c
roosmaa: Nothing specific. The Ledger wallet is completely un-opinionated about the path components and it's up to the wallet (or user) to choose the paths for the accounts. So the idea here was to be somewhat interoperable with the Ledger implementation, should the users want to use the extra path components.

keepkeyjon: Our attitude toward it is to strongly encourage one particular layout, and allow, but warn against using others. This encourages using standardized layouts since those are the ones with the nice human-readable descriptions, and gives others the raw bip32 path & warn on receive. It also allows people to migrate from nonstandard layouts if they need to, and helps users do the Right Thing (TM) that gives them compatibility with most other wallet implementations. Being too flexible here IMO leads to misplacement / apparent loss of funds, and difficult wallet UX.

roosmaa: Yeah, agreed.. it probably is the wrong place to introduce the flexibility.. I'll give this a do-over in the evening.

lib/firmware/app_layout.c
roosmaa: Fixed in b739a6b

lib/firmware/nano.c
roosmaa: Fixed in 958081b

lib/firmware/nano.c
roosmaa: Fixed in 958081b

keepkeyjon: ty!

lib/firmware/nano.c
```diff
@@ -0,0 +1,74 @@
...
+};
+
+bool nano_path_mismatched(const CoinType *coin,
+                          const uint32_t *address_n,
+                          const uint32_t address_n_count)
+{
+    // m/44' : BIP44-like path
+    // m / purpose' / bip44_account_path' / account'
+    bool mismatch = false;
+    mismatch |= address_n[0] != (0x80000000 + 44);
```
keepkeyjon:
```suggestion
    mismatch |= address_n_count > 0 && address_n[0] != (0x80000000 + 44);
```

roosmaa: Yes :man_facepalming: .. amended the commit

keepkeyjon:
I did some refactoring here: https://github.com/keepkey/keepkey-firmware/tree/feat/nano-support please have a look.

lib/firmware/fsm_msg_nano.h
```diff
@@ -0,0 +1,318 @@
...
+
+    memset(block_hash, 0, sizeof(block_hash));
+    memset(parent_hash, 0, sizeof(parent_hash));
+    memset(link, 0, sizeof(link));
+    memset(representative_pk, 0, sizeof(representative_pk));
+    memset(balance_be, 0, sizeof(balance_be));
+
+    // Calculate parent hash
+    if (msg->has_parent_block) {
+        if (msg->parent_block.has_parent_hash) {
```
keepkeyjon: AFAICT, this branch isn't exercised by any of the new python tests.

keepkeyjon:
I keep accidentally pushing to your fork instead of mine... sorry about that.

roosmaa:
I'll take a look at it on Sunday.

roosmaa:
Looks fine to me.

PR #84, opened 2019-02-11
Change Denarius Ticker to D

carsenk:
@keepkeyjon We changed our ticker last year from DNR to D, Also just curious to know if you need anything else from me to get Denarius implemented into the Keepkey? Thank you!


include/keepkey/firmware/coins.def
```diff
@@ -17,7 +17,7 @@ X(true, "EOS",          true, "EOS",   false, NA,    false,         0,  false, N
...
 X(true, "Fujicoin",     true, "FJC",   true,  36,    true,   10000000,  true,  16,  true, "\x19" "FujiCoin Signed Message:\n",      true, 0x8000004b,   false,   0,     true,   8,  false, NO_CONTRACT,      true, 76067358,  true, true,   true, false,  true, SECP256K1_STRING,  false, "",            true, "fc",    false, false,  true,  77429938,        true, 78792518)
-X(true, "Denarius",     true, "DNR",   true,  30,    true,     100000,  true,  90,  true, "\x19" "Denarius Signed Message:\n",      true, 0x80000074,   false,   0,     true,   8,  false, NO_CONTRACT,      true, 76067358,  true, false,  true, false,  true, SECP256K1_STRING,  false, "",            false, "",     false, false,  false,        0,        false,       0)
+X(true, "Denarius",     true, "D",   true,  30,    true,     100000,  true,  90,  true, "\x19" "Denarius Signed Message:\n",      true, 0x80000074,   false,   0,     true,   8,  false, NO_CONTRACT,      true, 76067358,  true, false,  true, false,  true, SECP256K1_STRING,  false, "",            false, "",     false, false,  false,        0,        false,       0)
```
keepkeyjon:
```suggestion
X(true, "Denarius",     true, "D",     true,  30,    true,     100000,  true,  90,  true, "\x19" "Denarius Signed Message:\n",      true, 0x80000074,   false,   0,     true,   8,  false, NO_CONTRACT,      true, 76067358,  true, false,  true, false,  true, SECP256K1_STRING,  false, "",            false, "",     false, false,  false,        0,        false,       0)
```

keepkeyjon: (aligns the columns in the table)

carsenk: Thanks for that, anymore insight on when it will actually be added to the Keepkey Client? Any other development that needs done?

keepkeyjon: I should set expectations here very low: having firmware support for a coin does not imply we'll be adding it to the KeepKey Client.

Best way to get it supported independent of that is to add KeepKey support via keepkey.js to some client that already supports Denarius.

carsenk: Ah, gotcha, well that is kind of a bummer. There is nothing that we can do as a community to attempt to get it in the Keepkey Client? I am assuming it is closed sourced?

carsenk: Also is Denarius available in the current stable firmware build for the Keepkey then?

keepkeyjon: It's in the beta firmware.

carsenk: Okay thanks, so keepkey.js has no client examples just the basic interfacing with the keepkey examples? Thanks for all the help!

keepkeyjon: MEW is the best open source example at the moment: https://github.com/MyEtherWallet/MyEtherWallet

There's also a simple xpub dumper here: https://github.com/keepkey/keepkey.js/pull/24#issuecomment-460348006

carsenk: Yea, one will just hope we are just added to your Keepkey Client at some point...I would be more than happy to commit development for it needed...

PR #87, opened 2019-02-26
Add Regtest CoinTable entry

### LINKED ISSUE #86: Support Bitcoin Regtest
It would be very useful for local development to have bitcoin regtest support in the coin table.

Here are the constants for regtest:
https://github.com/bcoin-org/bcoin/blob/f585d8616df8335fddbc21d1b776a979ec6ff305/lib/protocol/networks.js#L847

Satoshi Labs also has this slip:
https://github.com/satoshilabs/slips/blob/master/slip-0173.md

keepkeyjon:
Closes #86


keepkeyjon:
@tynes please review

tynes:
I went through the values and confirmed they match the expected values

bcoin doesn't currently use: 

`xpub_magic_segwit_p2sh`: 71979618 -> 0x044a5262 -> upub
`xpub_mmagic_segwit_native`: 73342198 -> 0x045f1cf6 -> vpub

but we have an open pull request to add `ypub` and `zpub` (the mainnet versions of upub and vpub), and we would eventually like to be widely compatible with everything, so its on the roadmap. It definitely makes sense to use these values because that it what bitcoin core uses and many other wallet services. we just use `m/44` for all kinds of addresses...

Thank you for the help here!

reference: https://github.com/satoshilabs/slips/blob/master/slip-0132.md

PR #112, opened 2019-04-19
webusb-friendly labelling

keepkeyjon:
image


keepkeyjon:
> The Zs were a nice touch. If you need copy, let me know and I can have Bethany spend a cycle on it

The "copy" here comes from the user's label, whatever they set it to, so long as it is ascii only.

PR #135, opened 2019-06-10
move strlcpy/strlcat to libboard

### LINKED ISSUE #134: Can't build KeepKey emulator (Ubuntu 18,  GCC 7)
I've failed to build KeepKey emulator.

What I'm doing:

`git clone https://github.com/keepkey/keepkey-firmware`
`cd keepkey-firmware`
`git submodule update --init --recursive`
`mkdir build`
`cd build`
`cmake -C ../cmake/caches/emulator.cmake ../ -DNANOPB_DIR=/path/to/nanopb-0.2.9.2 -DPROTOC_BINARY=/usr/bin/protoc`
`make`

Simple cmake output (command `cmake -C ../cmake/caches/emulator.cmake ../ -DNANOPB_DIR=home/sergey/src/nanopb-0.2.9.2 -DPROTOC_BINARY=/usr/bin/protoc >>cmake.out 2>&1`): https://pastebin.com/TAkT3nNa

CMake output in trace-expand mode (command `cmake --trace-expand -C ../cmake/caches/emulator.cmake ../ -DNANOPB_DIR=path/to/nanopb-0.2.9.2 -DPROTOC_BINARY=/usr/bin/protoc >>cmake.out 2>&1`): https://www.sendspace.com/file/gmxtve (sorry, it is too big, can't place it at pastebin)

make output (command `make >> make.out 2>&1`): https://pastebin.com/CijBjZDw

I'm not sure, but looks like `kkfirmware` lib with unresolved references is linked before `kkemulator` lib that contains implementations.

keepkeyjon:
Fixes #134


lib/board/CMakeLists.txt
```diff
@@ -22,6 +22,14 @@ set(sources
     util.c
     variant.c)
 
+if(NOT ${KK_HAVE_STRLCPY})
```
gladcow:
```suggestion
if(NOT DEFINED ${KK_HAVE_STRLCPY})
```

lib/board/CMakeLists.txt
```diff
@@ -22,6 +22,14 @@ set(sources
...
+    set(sources ${sources} strlcpy.c)
+endif()
+
+if(NOT ${KK_HAVE_STRLCAT})
```
gladcow:
```suggestion
if(NOT DEFINED ${KK_HAVE_STRLCAT})
```

lib/board/CMakeLists.txt
keepkeyjon: I don't think that's correct. CheckSymbolExists sets the variable to either `1` or `""` depending on whether it was found in the supplied header:

https://github.com/Kitware/CMake/blob/master/Modules/CheckSymbolExists.cmake#L124

And `if()` properly detects those two resulting values:

https://cmake.org/cmake/help/latest/command/if.html#condition-syntax

PR #141, opened 2019-06-18
firmware: put the device label on the home screen

keepkeyjon:


keepkeyjon:
erm, maybe not: 
image

PR #149, opened 2019-06-30
EOS: to maximize compatibilty, remove slip48, and go with ledger paths

keepkeyjon:
https://github.com/trezor/trezor-firmware/issues/168#issuecomment-495262620


keepkeyjon:
yeah. I still need to figure out why those are failing. this commit shouldn't have broken those tests.

keepkeyjon:
Turns out it was a combination of a bug and a bad test.

PR #176, opened 2019-10-13
firmware: be more strict about mixed-mode change

keepkeyjon:


keepkeyjon:
https://github.com/keepkey/python-keepkey/pull/83

PR #182, opened 2019-10-19
Remove word recovery

keepkeyjon:
It takes up space in .text/.data/.bss, and isn't being used.


keepkeyjon:
https://github.com/keepkey/python-keepkey/pull/80/files

PR #183, opened 2019-10-27
ShapeShift quick trades

keepkeyjon:


lib/firmware/app_confirm.c
```diff
@@ -138,19 +138,18 @@ bool confirm_decrypt_msg(const char *msg, const char *address)
...
+                             const char *wit_amt)
 {
-    return confirm_with_custom_layout(&layout_notification_no_title_bold,
+    return confirm_with_custom_layout(&layout_standard_notification,
                                       ButtonRequestType_ButtonRequest_SignExchange,
-                                      "",
-                                      "ShapeShift %s into\n%s and send to\n%s",
-                                      dep_amt, wit_amt, address);
+                                      "ShapeShift",
+                                      "Trade %s with ShapeShift in exchange for %s?",
```
SteveHadow: Do we need to add a `~` to show that its the estimated amount?

keepkeyjon: that is taken care of here: https://github.com/keepkey/keepkey-firmware/pull/183/files#diff-9fd5f39948911ad650f011be3248e530R594

PR #187, opened 2019-11-07
Add Polis

eabz:


greatwolf:
This PR should be rejected and closed since Polis has migrated to off their own blockchain and onto Binance Smart Chain. See https://polisblockchain.medium.com/polis-2-0-the-new-era-3832d9d92ce4

PR #188, opened 2019-11-12
Feature/cosmos

CaptJakk:


include/keepkey/firmware/stellar.h
```diff
@@ -0,0 +1,102 @@
+#ifndef __STELLAR_H__
```
keepkeyjon: Should rebase this & drop the stellar bits so they're in independent PRs.

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,82 @@
...
+#include "keepkey/board/confirm_sm.h"
+#include "keepkey/board/util.h"
+#include "keepkey/firmware/home_sm.h"
+#include "keepkey/firmware/storage.h"
+#include "trezor/crypto/segwit_addr.h"
+#include "trezor/crypto/ecdsa.h"
+#include <stdbool.h>
+#include <time.h>
+
+/* This was inlined from the trezor firmware because it was not exported.
```
keepkeyjon: Just make a PR to drop `static` from github.com/keepkey/trezor-firmware. We can live with a few small downstream changes here and there.

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,82 @@
...
+ * address: output buffer
+ *
+ * returns true if successful
+ */
+bool cosmos_getAddress(const uint8_t *public_key, char *address)
+{
+    uint8_t hash160Buf[RIPEMD160_DIGEST_LENGTH];
+    ecdsa_get_pubkeyhash(public_key, HASHER_SHA2_RIPEMD, hash160Buf);
+
+    uint8_t fiveBitExpanded[32]; // RIPEMD160_DIGEST_LENGTH * 8 / 5
```
keepkeyjon:
```suggestion
    uint8_t fiveBitExpanded[RIPEMD160_DIGEST_LENGTH * 8 / 5];
```

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,82 @@
...
+
+/*
+ * Gets the address
+ *
+ * public_key: 33 byte compressed secp256k1 key
+ * address: output buffer
+ *
+ * returns true if successful
+ */
+bool cosmos_getAddress(const uint8_t *public_key, char *address)
```
keepkeyjon: Would be better to take a `const HDNode *node`, and pull the pubkey from that.

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,82 @@
...
+                   const char* memo,
+                   const size_t memo_length,
+                   const uint64_t amount,
+                   const char* from_address,
+                   const char* to_address,
+                   const uint64_t sequence,
+                   uint8_t* signature)
+{
+    size_t len = 412 + memo_length + chain_id_length;
+    char signBytes[len];
```
keepkeyjon: I think we ought to avoid VLAs where possible, since they make it easier to write exploits that depend on specific stack alignment. Best way around that here would be to split up the `SIGNING_TEMPLATE`, and incrementally hash the components that go into it using `sha256_{Init,Update,Final}` instead of `sha256_Raw`.

Also, the `memo` field must be sanitized or escaped, since that would allow a malicious host to inject parts of transaction payload without showing the user what's really happening in the tx.

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    if (!confirm(ButtonRequestType_ButtonRequest_ProtectCall, _("Confirm Fee Details"), "Fee: %"PRIu32" uATOM\nGas: %"PRIu32"", msg->fee_amount, msg->gas))
+    {
+        fsm_sendFailure(FailureType_Failure_ActionCancelled, NULL);
+        layoutHome();
+        return;
+    }
+
+    if (!confirm(ButtonRequestType_ButtonRequest_ProtectCall, _("Confirm Aux Details"), "Memo: \"%s\"\nChain ID: %s\nAccount #: %"PRIu64"", msg->memo, msg->chain_id, msg->account_number))
```
keepkeyjon: I don't think we need to show the `chain_id`, nor the `account_number`, especially since the latter has a very confusing name for what it does in the Cosmos ecosystem.

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,94 @@
...
+{
+    CHECK_INITIALIZED
+    CHECK_PIN
+
+    HDNode *node = fsm_getDerivedNode(SECP256K1_NAME, msg->address_n, msg->address_n_count, NULL);
+    if (!node) { return; }
+    hdnode_fill_public_key(node);
+
+    // Confirm transaction basics
+    if (!confirm(ButtonRequestType_ButtonRequest_ProtectCall, _("Confirm Main Details"), "From: %s\nTo: %s\nAmount: %f ATOM", msg->from_address, msg->to_address, (float)msg->amount * 1E-6))
```
keepkeyjon: It would be more user-friendly to show the account description than the `from_address`. For example: https://github.com/keepkey/keepkey-firmware/blob/695cbdc8fe04298179936dab0f81bf6a8701814a/lib/firmware/fsm_msg_coin.h#L217

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,94 @@
...
+    CHECK_INITIALIZED
+
+    CHECK_PIN
+
+    HDNode *node = fsm_getDerivedNode(SECP256K1_NAME, msg->address_n, msg->address_n_count, NULL);
+    if (!node) { return; }
+    hdnode_fill_public_key(node);
+
+    cosmos_getAddress(node->public_key, resp->address);
+
```
keepkeyjon: needs one of these: https://github.com/keepkey/keepkey-firmware/blob/695cbdc8fe04298179936dab0f81bf6a8701814a/lib/firmware/fsm_msg_coin.h#L229

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,94 @@
...
+    if (msg->has_show_display && msg->show_display)
+    {
+        if (!confirm(ButtonRequestType_ButtonRequest_ProtectCall, _("Share public account ID?"), "%s", resp->address))
```
keepkeyjon: Would be better to use `confirm_address` for this: https://github.com/keepkey/keepkey-firmware/blob/695cbdc8fe04298179936dab0f81bf6a8701814a/lib/firmware/fsm_msg_coin.h#L240

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,94 @@
+#include "keepkey/firmware/cosmos.h"
```
keepkeyjon: this include should go in `fsm.c`

lib/firmware/fsm_msg_cosmos.h
keepkeyjon: Please add the account description like we have here:

https://github.com/keepkey/keepkey-firmware/blob/695cbdc8fe04298179936dab0f81bf6a8701814a/lib/firmware/fsm_msg_coin.h#L217-L223

lib/firmware/cosmos.c
CaptJakk: Should escaping happen on ALL %s replacements then?

keepkeyjon: Either that, or make sure that they're properly escaped by design. For example, hypothetically, say the `chain_id` is always a hex string, then you could make the protocol take bytes, and emit them as hex into the hash, and that wouldn't need escaping.

lib/firmware/cosmos.c
CaptJakk: Should this change be made to master?

keepkeyjon: The PR should be against the `keepkey` branch. `master` reflects upstream.

lib/firmware/coins.c
```diff
@@ -432,6 +432,13 @@ static bool isEOS(const char *coin_name) {
     return false;
 }
 
+static bool isAccountBased(const char* coin_name)
+{
+    if (strcmp(coin_name, "Cosmos") == 0) { return true; }
```
keepkeyjon: If you put `EOS` in here too, it simplifies the two call sites a lot.

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,107 @@
...
+                            const uint32_t address_n_count)
+{
+    // m/44' : BIP44-like path
+    // m / purpose' / bip44_account_path' / account' / x / y
+    bool mismatch = false;
+    mismatch |= address_n_count != 5;
+    mismatch |= address_n_count > 0 && (address_n[0] != (0x80000000 + 44));
+    mismatch |= address_n_count > 1 && (address_n[1] != _coin->bip44_account_path);
+    mismatch |= address_n_count > 2 && (address_n[2] & 0x80000000) == 0;
+    mismatch |= address_n_count > 2 && (address_n[3] & 0x80000000) == 0;
```
keepkeyjon: undefined behavior
```suggestion
    mismatch |= address_n_count > 3 && (address_n[3] & 0x80000000) == 0;
```

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,107 @@
...
+    mismatch |= address_n_count > 2 && (address_n[4] & 0x80000000) == 0;
```
keepkeyjon:
```suggestion
    mismatch |= address_n_count > 4 && (address_n[4] & 0x80000000) == 0;
```

lib/firmware/cosmos.c
keepkeyjon: also, x and y shouldn't be hardened. maybe you meant `!= 0`?

keepkeyjon: or are they supposed to be zero, ie:
```
mismatch |= address_n_count > 3 && address_n[3] != 0;
```

CaptJakk: great catch, no they need to not be hardened.

lib/firmware/coins.c
```diff
@@ -448,7 +456,7 @@ bool bip32_node_to_string(char *node_str, size_t len, const CoinType *coin,
             return false;
 
         // Only 0/1 for internal/external are valid paths on UTXO coins.
-        if (!isEthereumLike(coin_name) && !isEOS(coin_name) &&
+        if (!isEthereumLike(coin_name) && !isEOS(coin_name) && !isAccountBased(coin_name)
```
keepkeyjon:
```suggestion
        if (!isAccountBased(coin_name) &&
```

lib/firmware/coins.c
```diff
@@ -460,7 +468,7 @@ bool bip32_node_to_string(char *node_str, size_t len, const CoinType *coin,
     if (!prefix)
         return false;
 
-    if (whole_account || isEthereumLike(coin_name) || isEOS(coin_name) || !show_addridx) {
+    if (whole_account || isEthereumLike(coin_name) || isEOS(coin_name) || isAccountBased(coin_name) || !show_addridx) {
```
keepkeyjon:
```suggestion
    if (whole_account || isAccountBased(coin_name) || !show_addridx) {
```

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,107 @@
...
+                            const uint32_t address_n_count)
+{
+    // m/44' : BIP44-like path
+    // m / purpose' / bip44_account_path' / account' / x / y
+    bool mismatch = false;
+    mismatch |= address_n_count != 5;
+    mismatch |= address_n_count > 0 && (address_n[0] != (0x80000000 + 44));
+    mismatch |= address_n_count > 1 && (address_n[1] != _coin->bip44_account_path);
+    mismatch |= address_n_count > 2 && (address_n[2] & 0x80000000) == 0;
+    mismatch |= address_n_count > 3 && (address_n[3] & 0x80000000) != 0;
```
keepkeyjon: If most other cosmos wallets use `m/44'/118'/account'/0/0` then this should be:

```suggestion
    mismatch |= address_n_count > 3 && address_n[3]  != 0;
```

And we should set show_addr_idx=false, whole_account=true like we do for ETH in the `node_str` calculations.

lib/firmware/cosmos.c
CaptJakk: I'm running into the issue now where it's upset about a missing ed25519-blake2b.c source file in the CMakeLists.txt at this point, so somewhere between the locked commit, and keepkey/master, this file was removed, what should we do here? can we remove that file from the CMakeLists? I think it's out of scope to fix build issues unrelated to the cosmos update.

CaptJakk: OK, so I tried just removing it but this is a thread that is unraveling a lot right now.

keepkeyjon: I'll have a look.

keepkeyjon: Can you try with this commit? https://github.com/keepkey/trezor-firmware/pull/1/files

keepkeyjon: > Clearly keepkey/trezor-firmware/master has a lot of incompatible changes with the current branch of keepkey/keepkey-firmware/master. Please advise

`keepkey-firmware` is using `keepkey/trezor-firmware/keepkey` not` keepkey/trezor-firmware/master`, hence why I asked you to make the commit against the `keepkey` branch, not `master`.

CaptJakk: 🤦‍♂ my bad. Thanks.

keepkeyjon: :)

CaptJakk: was `mnemonic_wordlist` removed in this fast forward? I'm getting complaints about implicit declarations and I don't recall changing any of that code

keepkeyjon: bee9b60e4ce0a48d62494148fd62344491e3b2a6 deals with that. upstream renamed it to `wordlist`, IIRC.

keepkeyjon: you probably just need to merge latest `keepkey-firmware/master` into your branch.

CaptJakk:
OK. So we refactored it to go through the message protocol you asked, and the python-keepkey, and device-protocol PR's should reflect that. Everything has been run through the build and tests as is.

include/keepkey/firmware/cosmos.h
```diff
@@ -0,0 +1,51 @@
...
+bool cosmos_signTxUpdateMsgSend(const uint64_t amount,
+                                const char *from_address,
+                                const char *to_address);
+bool cosmos_signTxFinalize(uint8_t* public_key, uint8_t* signature);
+bool cosmos_signingIsInited(void);
+bool cosmos_signingIsFinished(void);
+void cosmos_signAbort(void);
+size_t cosmos_getAddressNCount(void);
+bool cosmos_getAddressN(uint32_t* address_n, size_t address_n_count);
+// bool cosmos_signTx(const uint8_t* private_key,
```
keepkeyjon: rm dead comment

include/keepkey/firmware/cosmos.h
```diff
@@ -0,0 +1,51 @@
...
+#include "fsm.h"
+#include "messages.pb.h"
+#include "trezor/crypto/bip32.h"
+
+bool cosmos_path_mismatched(const CoinType *_coin,
+                            const uint32_t *_address_n,
+                            const uint32_t _address_n_count);
+bool cosmos_getAddress(const HDNode *_node, char *address);
+bool cosmos_signTxInit(const HDNode* _node,
+                       const uint32_t _address_n[8],
```
keepkeyjon: The `8` isn't enforced unless you do `[static 8]`. Better to make it a `const uint32_t *_address_n`, and let `address_n_count` do its thing.

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,307 @@
...
+                            const uint32_t _address_n_count)
+{
+    // m/44' : BIP44-like path
+    // m / purpose' / bip44_account_path' / account' / x / y
+    bool mismatch = false;
+    mismatch |= _address_n_count != 5;
+    mismatch |= _address_n_count > 0 && (_address_n[0] != (0x80000000 + 44));
+    mismatch |= _address_n_count > 1 && (_address_n[1] != _coin->bip44_account_path);
+    mismatch |= _address_n_count > 2 && (_address_n[2] & 0x80000000) == 0;
+    mismatch |= _address_n_count > 3 && (_address_n[3] & 0x80000000) != 0;
```
keepkeyjon:
```suggestion
    mismatch |= _address_n_count > 3 && _address_n[3]  != 0;
```

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,307 @@
...
+    mismatch |= _address_n_count > 4 && (_address_n[4] & 0x80000000) != 0;
```
keepkeyjon:
```suggestion
    mismatch |= _address_n_count > 4 && _address_n[4] != 0;
```

lib/firmware/cosmos.c
keepkeyjon: If we leave this too flexible, there's a chance of ransom attacks like what was disclosed in https://thecharlatan.github.io/Ransom-Coldcard/

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,307 @@
...
+ */
+bool cosmos_getAddress(const HDNode *_node, char *address)
+{
+    uint8_t hash160Buf[RIPEMD160_DIGEST_LENGTH];
+    ecdsa_get_pubkeyhash(_node->public_key, HASHER_SHA2_RIPEMD, hash160Buf);
+
+    uint8_t fiveBitExpanded[RIPEMD160_DIGEST_LENGTH * 8 / 5];
+    size_t len = 0;
+    convert_bits(fiveBitExpanded, &len, 5, hash160Buf, 20, 8, 1);
+    // bech32encode
```
keepkeyjon:
```suggestion
```

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,307 @@
...
+    return bech32_encode(address, "cosmos", fiveBitExpanded, len) == 1;
+}
+
+// Each segment guaranteed to be less than or equal to 64 bytes
+// 19 + ^20 + 14 = ^53
+#define SIGNING_TEMPLATE_START_SEG1 "{\"account_number\":\"%" PRIu64 "\",\"chain_id\":\""
```
keepkeyjon: I think these should all be inlined where they're used. The macros aren't buying us much, and they arguably make it less clear what's going on.

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,307 @@
...
+                                const char *from_address,
+                                const char *to_address)
+{
+    int n;
+    char buffer[SHA256_BLOCK_LENGTH + 1]; // NULL TERMINATOR NOT PART OF HASH
+
+    size_t decoded_len;
+    char hrp[45];
+    uint8_t decoded[38];
+    if (!bech32_decode(hrp, decoded, &decoded_len, from_address)) { return false; }
```
keepkeyjon: Can the `from_address` be derived from the signer's pubkey? When would it ever not be the same? If so, we should do address encoding here, instead of decode & check.

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,307 @@
...
+//     if (n < 0)
+//     {
+//         return false;
+//     }
+//     sha256_Update(&ctx, (uint8_t *)buffer, n);
+
+//     uint8_t hash[SHA256_DIGEST_LENGTH];
+//     sha256_Final(&ctx, hash);
+//     return ecdsa_sign_digest(&secp256k1, private_key, hash, signature, NULL, NULL) == 0;
+// }
```
keepkeyjon: rm dead code

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,182 @@
...
+    hdnode_fill_public_key(node);
+
+    if (!confirm(ButtonRequestType_ButtonRequest_ProtectCall, _("Confirm Fee Details"), "Fee: %" PRIu32 " uATOM\nGas: %" PRIu32 "", msg->fee_amount, msg->gas))
+    {
+        fsm_sendFailure(FailureType_Failure_ActionCancelled, NULL);
+        layoutHome();
+        return;
+    }
+
+    if (!confirm(ButtonRequestType_ButtonRequest_ProtectCall, _("Confirm Aux Details"), "Memo: \"%s\"\nChain ID: %s", msg->memo, msg->chain_id))
```
keepkeyjon:
```suggestion
    if (msg->has_memo && !confirm(ButtonRequestType_ButtonRequest_ProtectCall, _("Confirm Aux Details"), "Memo: \"%s\"\nChain ID: %s", msg->memo, msg->chain_id))
```

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,182 @@
...
+        return;
+    }
+    size_t address_n_count = cosmos_getAddressNCount();
+    char node_str[NODE_STRING_LENGTH];
+    if (!bip32_node_to_string(node_str, sizeof(node_str), coin, address_n,
+                              address_n_count, /*whole_account=*/false,
+                              /*show_addridx=*/true) &&
+        !bip32_path_to_string(node_str, sizeof(node_str),
+                              address_n, address_n_count)) {
+        memset(node_str, 0, sizeof(node_str));
```
keepkeyjon: I think we need to `fsm_sendFailure` after the memset, since otherwise there isn't a human-readable account description in `node_str`.

lib/firmware/cosmos.c
CaptJakk: is the intent here to only ever produce `m/44'/118'/0'/0/0` ?

lib/firmware/cosmos.c
CaptJakk: I'm fine doing that, but the reason we went this route is because the format strings have to either be literals or macros. If we inline them it adds a bit of noise to the line and we had opted to extract that noise to a single place. I agree that the *_SEG# is somewhat arbitrary and for that reason it can be very unclear. However, it was important that all of the strings with their maximum expansions for the wildcards fell below the buffer size.

lib/firmware/cosmos.c
CaptJakk: Yes we can do that

keepkeyjon: ok, in that case, it's fine as-is.

lib/firmware/cosmos.c
keepkeyjon: `m/44'/118'/account'/0/0`

lib/firmware/cosmos.c
```diff
@@ -0,0 +1,225 @@
...
+
+    initialized = true;
+    msgs_remaining = msg_count;
+    has_message = false;
+    sequence = _sequence;
+
+    memzero(&node, sizeof(node));
+    memcpy(&node, _node, sizeof(node));
+    memzero(address_n, sizeof(address_n));
+    memcpy(address_n, _address_n, sizeof(address_n));
```
keepkeyjon: Potential for UB here, if `address_n` and `_address_n` aren't the same length. I recommend:

checking that:

```
_address_n_count <= sizeof(address_n)/sizeof(address_n[0])
```

and:

```suggestion
    memcpy(address_n, _address_n, _address_n_count * sizeof(*address_n));
```

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    if (!confirm(ButtonRequestType_ButtonRequest_ProtectCall, _("Confirm Send Details"), "From: %s\nTo: %s\nAmount: %f ATOM", node_str, msg->send.to_address, (float)msg->send.amount * 1E-6))
```
keepkeyjon: Float math isn't appropriate displaying send amounts. Have a look at `bn_format_uint64` from `trezor/crypto/bignum.h`.

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,184 @@
...
+        RESP_INIT(CosmosMsgRequest);
+        msg_write(MessageType_MessageType_CosmosMsgRequest, resp);
+        return;
+    }
+
+    RESP_INIT(CosmosSignedTx);
+
+    if(!cosmos_signTxFinalize(resp->public_key.bytes, resp->signature.bytes)) {
+        fsm_sendFailure(FailureType_Failure_FirmwareError, "Failed to finalize signature");
+        layoutHome();
```
keepkeyjon:
```suggestion
        cosmos_signAbort();
        layoutHome();
```

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,184 @@
...
+        return;
+    }
+
+    resp->public_key.size = 33;
+    resp->has_public_key = true;
+    resp->signature.size = 64;
+    resp->has_signature = true;
+    layoutHome();
```
keepkeyjon:
```suggestion
        cosmos_signAbort();
        layoutHome();
```

lib/firmware/fsm_msg_cosmos.h
```diff
@@ -0,0 +1,184 @@
...
+
+    if(!cosmos_signTxUpdateMsgSend(msg->send.amount, msg->send.to_address)) {
+        cosmos_signAbort();
+        fsm_sendFailure(FailureType_Failure_FirmwareError, "Failed to include send message in transaction");
+        return;
+    }
+
+    if (!cosmos_signingIsFinished()) {
+        RESP_INIT(CosmosMsgRequest);
+        msg_write(MessageType_MessageType_CosmosMsgRequest, resp);
```
keepkeyjon:
```suggestion
        msg_write(MessageType_MessageType_CosmosMsgRequest, resp);
        cosmos_signAbort();
```

keepkeyjon:
I made some changes here that are easier to explain with code than in prose: https://github.com/Start9Labs/keepkey-firmware/pull/2

keepkeyjon:
@CaptJakk more little things: https://github.com/Start9Labs/keepkey-firmware/pull/3/files

CaptJakk:
What's the status on this?

lib/firmware/fsm_msg_cosmos.h
keepkeyjon: Still need this one to clear the state of the signing state machine.

keepkeyjon:
> What's the status on this?

Looks generally good. I have a small PR against the python-keepkey changes https://github.com/Start9Labs/python-keepkey/pull/1/files, and after that I need to run the tests and do a few on-chain txs from a real device as sanity check. If that checks out, this should be good to merge.

CaptJakk:
I just committed the state clearing and merged the python tests PR.

CaptJakk:
just wanted to check in on this. I saw they python and device protocol merges happen. Is there anything left on this besides some device testing on your part?

keepkeyjon:
> Is there anything left on this besides some device testing on your part?

Just testing on my part for now.

PR #209, opened 2020-01-09
Drop support for KeepKey Client style ERC20 sends

keepkeyjon:


keepkeyjon:
It is expected that this breaks this test: https://github.com/keepkey/python-keepkey/blob/2170fc4479bf5a4b63d2067801ecf041430db769/tests/test_msg_ethereum_erc20_signtx_exchange.py#L58

We'll delete that once this gets merged.

PR #210, opened 2020-01-13
Improve Recovery UX with has_word_count

keepkeyjon:
If word_count is set, finalize the recovery process if the user hits ' ' after
the last word, instead of complaining that they've entered too many.


keepkeyjon:
https://github.com/keepkey/python-keepkey/pull/86#discussion_r365888233

keepkeyjon:
https://github.com/keepkey/python-keepkey/pull/87

Issue #216, opened 2020-01-18
Emulator build


antonio-fr:
Can't make the emulator build at all in Ubuntu18.04 :

First the[ doc build ](https://github.com/keepkey/keepkey-firmware/blob/v6.3.0/docs/Build.md) is so outdated and so useless. This now requires a recursive git clone, and the given build directory fails.

```
cd ~
wget https://jpa.kapsi.fi/nanopb/download/nanopb-0.2.9.2.tar.gz
tar xzf nanopb-0.2.9.2.tar.gz
cd nanopb/generator/proto
make
export PATH=~/nanopb/generator:$PATH
cd ~
git clone https://github.com/keepkey/keepkey-firmware
cd keepkey-firmware
git checkout v6.3.0
git submodule update --init --recursive
mkdir build
cd build
cmake -C ../cmake/caches/emulator.cmake ../
make
```

I get the error message :
```
[ 98%] Building CXX object unittests/firmware/CMakeFiles/firmware-unit.dir/storage.cpp.o
/home/moa/keepkey-firmware/unittests/firmware/storage.cpp: In member function ‘virtual void Storage_WriteMeta_Test::TestBody()’:
/home/moa/keepkey-firmware/unittests/firmware/storage.cpp:38:5: error: C99 designator ‘uuid’ outside aggregate initializer
     };
     ^
/home/moa/keepkey-firmware/unittests/firmware/storage.cpp:38:5: error: C99 designator ‘uuid_str’ outside aggregate initializer
/home/moa/keepkey-firmware/unittests/firmware/storage.cpp: In member function ‘virtual void Storage_WriteCacheV1_Test::TestBody()’:
/home/moa/keepkey-firmware/unittests/firmware/storage.cpp:188:5: error: C99 designator ‘root_seed_cache’ outside aggregate initializer
     };
     ^
/home/moa/keepkey-firmware/unittests/firmware/storage.cpp: In member function ‘virtual void Storage_DumpNode_Test::TestBody()’:
/home/moa/keepkey-firmware/unittests/firmware/storage.cpp:229:5: sorry, unimplemented: non-trivial designated initializers not supported
     };
     ^
/home/moa/keepkey-firmware/unittests/firmware/storage.cpp: In member function ‘virtual void Storage_UpgradePolicies_Test::TestBody()’:
/home/moa/keepkey-firmware/unittests/firmware/storage.cpp:589:5: error: C99 designator ‘policy_name’ outside aggregate initializer
     };
[...]

PR #217, opened 2020-01-22
unittests: don't rely on C99 designator extensions from clang

keepkeyjon:


keepkeyjon:
Fixes the build errors from #216.

PR #181, opened 2019-10-18
Stellar Support

ProofOfKeags:
This change adds support for stellar "get_address" and "sign_tx" over the fsm, as well as all of the required operations in order to be able to accomplish that.

Note: the current release build overruns the 1MB flash by 16K, in theory theres enough space for the bootstrap, bootloader, and firmware, but it will require compressing the space between the bootstrap and bootloader images by a little bit.

PR #221, opened 2020-02-03
firmware: Stellar support

keepkeyjon:


keepkeyjon:
need to fix all the things here: https://github.com/keepkey/keepkey-firmware/pull/181

PR #225, opened 2020-02-13
firmware: fix U2F dialog UX in a better way

keepkeyjon:
... for the case where the host is requesting signature from a different U2F
device. Instead of showing the request for registration dialog as we currently
do, it seems better to display nothing at all.


keepkeyjon:
> Are you leaving the #if 1 and if so, why?

To make it clear what's missing there.

PR #238, opened 2020-11-22
adds bep2 support for binance

ProofOfKeags:
OK so there really wasn't that much to do. All we needed to do was remove the guard rails and then format the denom on the screen correctly rather than hard coding it. The existing protobuf spec was sufficient to be able to handle this.


markrypto (changes requested):
Please push a feature branch to master and create a new PR to merge that branch. Also, please certify that you have tested your new feature on a keepkey and it behaves as expected prior to requesting a merge into the master branch.

ProofOfKeags:
I'm not sure what you mean by "push a feature branch to master". The code is on a feature branch of our fork and we do not have the ability to push anything directly upstream. There is also a test that is in python-keepkey that verifies the functionality in the changes here.

markrypto (approved):
Sorry, I didn't get the pr notification from the python repo but I see your test there, thanks.

Everything looks good.

PR #248, opened 2021-02-23
Rel v6.7.0

markrypto:


mcchadwick (approved):

mvpratt (approved):
approved with comments

mvpratt:
why does this PR show commit history back to 2019, from previous releases, in the diff?

mvpratt:
@reidrankin can you elaborate?  I don't see any key derivation changes in this PR

reidrankin:
> @reidrankin can you elaborate?  I don't see any key derivation changes in this PR

That comment was on commit cc378f2, which reduces PIN KDF iterations by a factor of 10,000.

reidrankin:
To clarify, the changes were reverted in 430b404. That's also a prt of this PR so it's a wash; I'm just interested in the history behind those two changes. I'm confused because 430b404 says the reversion was for testing purposes -- whereas it would seem more likely to me that the initial reduction was for testing. (And besides, I'd expect you'd want to ease testing by reducing the values of the debug build defines, but not the normal ones!)

PR #251, opened 2021-03-04
Added Reddcoin (RDD) support

greatwolf:
This PR contains the changes necessary for adding Reddcoin support on the KeepKey hardware wallet.

### Explanation of code changes:
 - Added appropriate definition for Reddcoin in coins.def.
 - Reddcoin rawtransactions has an extra `timestamp`(UNIX little-endian) field appended at the end during serialization. That field has been added to the relevant code parts and changes made to accommodate that.
 - This PR needs to be complimented by keepkey/device-protocol#43 for `timestamp` field support in protobufs definition.
 - During the signing process for a rdd transaction, the `[SignTx]` and `[PrevTxMeta]` message needs to include a non-zero `uint32 timestamp` value from the host. Omitting this field(or `0` value) means `timestamp` won't be serialized as part of final rawtransaction. It's done this way to keep compatibility with other btc-style transactions like Bitcoin Cash, Dogecoin, Dash etc. since that field is always omitted.

Note about reddcoin transaction. The `timestamp` *does not* participate in the prehash that's being signed but it *does* participate in the final txid hash. This means for a reddcoin transaction, the `timestamp` could be modified without invalidating the signature but will change the txid.

Also note most of the changes involving `timestamp` are just backported from https://github.com/keepkey/trezor-firmware/blob/f4e6a4ed2a4f0bb7cf82b4e143f16a73e7845779/legacy/firmware/signing.c but adapted to work with reddcoin.


TechAdeptRDD:
Please advise why "Closed, not selected"?

PR #278, opened 2021-07-07
some fixes for #277

reidrankin:
This is a rollup of the suggestions in my review on #277; I'll copy the comments over as appropriate.

PR #279, opened 2021-07-07
some fixes for #278

reidrankin:
(This is the same as #278, except for 7d6bfe0, bcab1c3, and as otherwise noted.)

### DIFF
--- a/include/keepkey/board/pubkeys.h
+++ b/include/keepkey/board/pubkeys.h
@@ -23,7 +23,7 @@
 #include <inttypes.h>
 
 #define PUBKEYS 5
-#define EXP_PUBKEYS 5
+#define BLK_v2_0_0_PUBKEYS 5
 #define PUBKEY_LENGTH 65
 #define SIGNATURES 3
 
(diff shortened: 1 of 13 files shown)


lib/firmware/fsm_msg_tendermint.h
```diff
@@ -154,7 +154,10 @@ void fsm_msgTendermintMsgAck(const TendermintMsgAck *msg) {
     case OutputAddressType_TRANSFER:
     default: {
       char amount_str[32];
-      bn_format_uint64(msg->send.amount, NULL, msg->denom, 6, 0, false, amount_str,
+      char suffix[sizeof(coin->coin_shortcut) + 1]; // sizeof(coin->coin_shortcut) includes space for the terminator
+      strlcpy(suffix, " ", sizeof(suffix));
+      strlcat(suffix, coin->coin_shortcut, sizeof(suffix));
```
reidrankin, line 159: changed from #278: use `strlcat` instead of `snprintf`; slightly lighter, probably slightly safer

reidrankin (commented):
(submodule was also updated to point to a signed commit)

include/keepkey/board/pubkeys.h
```diff
@@ -23,7 +23,7 @@
 #include <inttypes.h>
 
 #define PUBKEYS 5
-#define EXP_PUBKEYS 5
+#define BLK_v2_0_0_PUBKEYS 5
```
reidrankin, line 26: (note that EXP_PUBKEYS is unused -- hence the overwrite)

markrypto (approved):

Issue #281, opened 2021-07-10
[CRITICAL ISSUE] Building latest 5ab76aa results in a broken firmware binary.


greatwolf:
I'm running into an issue where building the latest v7.1.5 5ab76aa results in a non-functional firmware. I'm able to upload onto my keepkey fine but afterwards my keepkey would be stuck on startup becoming non-responsive. The animated KeepKey logo never shows up nor does the `get/started` welcome screen.

The warning of unofficial firmware does come up however. After pressing the top button and confirming that is where my keepkey just stays in a non-responsive state.

Steps to reproduce:
 - git clone to the latest tip of the repo
 - recursively init and update all the dependent submodules as needed for current tip
 - cd into `./scripts/build/docker/device`
 - run `./release.sh` build script
 - after build completes upload the the produced firmware binary file `./bin/firmware.keepkey.bin` onto keepkey device.
 
I've tried diagnosing the issue and so far I've narrowed it down to some commit introducing this between bc40158 and 852f8ac.
Commit 852f8ac works fine btw, I am able to produce a working firmware from that.

Other thing to note, v7.1.5 binary in the release section seems to also work and that is tagged to the commit 5ab76aa. Again, manually building 5ab76aa myself produces a broken firmware; same as bc40158. How were you guys able to produce a functioning firmware from that? Was a different build process used instead of the `./release.sh` and `./debug.sh` scripts?

greatwolf:
After some more investigation and debugging I've narrowed down the problem to commit 0527170. Specifically the problematic [line is](https://github.com/keepkey/keepkey-firmware/commit/05271703b8c1863262f6df0d089d273fc2749ac8#diff-0b10186cddda4783b1e93def8cdd72dd91e0e42b4cc411fbfacf1d09c1fb7be8R97):

    return !fi_defense_delay(sigPresent);

In the original line before this commit you have:

    return *((uint8_t *)FLASH_META_SIGINDEX1) != 0;

I haven't looked inside `fi_defense_delay` but assuming that function just passes through `sigPresent` as a return or converts it to a `bool` then to keep the semantics of the original piece of code it should be something like this instead:

    return fi_defense_delay(sigPresent) != 0;
    // or simply
    return fi_defense_delay(sigPresent);

Changing that line and recompiling produces a working firmware again. My KeepKey no longer seizes up and becomes unresponsive.

markrypto:
greatwolf, what bootloader are you using?

reidrankin:
[...]

PR #282, opened 2021-07-12
fix backwards logic in canDropPrivs()

reidrankin:
0527170 contained an extra `!` which, when using bootloader v2.0.0, breaks unsigned firmware and prevents signed builds from dropping their privs.

This error made it past our testing regime because it didn't trigger on our normal testing devices; props to @greatwolf for bringing it to our attention as #281.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.7.2)
 
 project(KeepKeyFirmware
 
-        VERSION 7.1.5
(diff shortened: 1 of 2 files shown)


BitHighlander (dismissed):
🤦

markrypto (approved):

reidrankin:
@markrypto (please make sure to reset master to 04ef714 instead of adding a merge commit, since the v7.1.6 binary has that commit hash embedded)

mvpratt (dismissed):

PR #285, opened 2021-07-13
Add downgrade protection

mrnerdhair:
This adds a form of downgrade protection which prevents as-yet-undiscovered vulnerabilities in older firmware versions from being used to compromise the keys of a user who has upgraded, but whose device is subsequently stolen by an attacker who downgrades it to a vulnerable version.

Storage format 17 is introduced, which is identical to format 16 except that the three bytes at offsets 1, 2, and 3 are used to store the major, minor, and patch version numbers of the firmware version which wrote the structure. Moving to a newer firmware version will cause these version numbers to be updated; moving to an older version will cause the storage to be reset and the user's private keys to be wiped -- the same thing that would happen if an unsigned firmware image were loaded.

A warning is displayed before resetting the storage, which provides a user the opportunity to unplug the device and update it to avoid the wipe. However, if the user chooses to downgrade to a version released before this warning message was added, their keys will be wiped without any warning beyond the bootloader's usual admonishment that your keys might be erased after an upgrade and that you should verify that your recovery sentence is available.

This approach increases security while maintaining a user's sovereignty over what firmware runs on their device.


markrypto (changes requested):
This is not the correct approach:

When adding params to the storage section, add them to the reserve and decrease the reserve accordingly. Use the storage_write/readv11 to read/write.

docs/Storage.md
```diff
@@ -167,7 +167,7 @@ STORAGE_VERSION 16 layout
 | u2f_counter               | u32            |            4 |            401 |
 | sec_fingerprint           | char[32]       |           32 |            405 |
 | random_salt               | char[32]       |           32 |            437 |
-| reserved                  | char[1028]     |         1028 |            469 |
+| reserved                  | char[1025]     |         1028 |            469 |
```
reidrankin, line 170:
```suggestion
| reserved                  | char[1028]     |         1028 |            469 |
```

unittests/firmware/storage.cpp
```diff
@@ -683,7 +686,7 @@ TEST(Storage, StorageRoundTrip) {
       0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
       0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
       0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-      0x00, 0x00, 0x00, 0x00, 0x00, 0x10, 0x00, 0x00, 0x00, 0xe4, 0x8d, 0xfe,
+      0x00, 0x00, 0x00, 0x00, 0x00, 0x11, 0x00, 0x00, 0x00, 0xe4, 0x8d, 0xfe,
```
reidrankin, line 689: encrypted_secrets_version

reidrankin:
(just for posterity, in case it's not clear from the history above, @markrypto's feedback about storage format has been addressed.)

markrypto (dismissed):

reidrankin:
Are we OK to move forward with this now that the language is approved?

markrypt0:
Closed, not selected.

PR #287, opened 2021-07-19
directly check priv mode before attempting to change

markrypto:


mcchadwick (approved):

reidrankin:
Very nice. I'm wondering if we even need the switch statement at all any more; I think that logic will work regardless of bootloader  version. I'm planning to do a test build tomorrow and try it against several bootloader versions.

markrypto:
I just checked the code and we do initialize the PSP stack in startup.s, so I believe we can remove that switch statement.

PR #288, opened 2021-07-22
Add support for EIP-1559 gas format

mcchadwick:


lib/firmware/ethereum.c
```diff
@@ -541,7 +583,11 @@ static void layoutEthereumFee(const uint8_t *value, uint32_t value_len,
  */
 
 static bool ethereum_signing_check(EthereumSignTx *msg) {
-  if (!msg->has_gas_price || !msg->has_gas_limit) {
+  if (!msg->has_gas_limit) {
+    return false;
+  }
+
+  if (!(msg->has_gas_price || msg->has_max_fee_per_gas)) {
```
markrypto, line 590: As I understand this logic, a tx has to have has_gas_price or has_max_fee_per_gas or both?

lib/firmware/ethereum.c
```diff
@@ -487,28 +487,70 @@ static void layoutEthereumData(const uint8_t *data, uint32_t len,
...
+  bn_zero(fee);
+
+  memset(pad_val, 0, sizeof(pad_val));
+  memcpy(pad_val + (32 - max_fee_per_gas_len), max_fee_per_gas,
+         max_fee_per_gas_len);
+  bn_read_be(pad_val, &max_fee);
+
+  memset(pad_val, 0, sizeof(pad_val));
+  memcpy(pad_val + (32 - max_priority_fee_per_gas_len),
+         max_priority_fee_per_gas, max_priority_fee_per_gas_len);
```
markrypto, line 516: Same as above

lib/firmware/ethereum.c
```diff
@@ -724,10 +770,13 @@ void ethereum_signing_init(EthereumSignTx *msg, const HDNode *node,
...
-                    msg->gas_price.size, msg->gas_limit.bytes,
-                    msg->gas_limit.size, token != NULL, confirm_body_message,
-                    sizeof(confirm_body_message));
+  layoutEthereumFee(
+      msg->value.bytes, msg->value.size, msg->gas_price.bytes,
+      msg->gas_price.size, msg->gas_limit.bytes, msg->gas_limit.size,
+      msg->max_fee_per_gas.bytes, msg->max_fee_per_gas.size,
+      msg->max_priority_fee_per_gas.bytes, msg->max_priority_fee_per_gas.size,
+      msg->has_max_fee_per_gas ? true : false, token != NULL,
+      confirm_body_message, sizeof(confirm_body_message));
```
markrypto, line 779: Wow, big param list. Is it better just to pass msg rather than a list of everything in msg here?

markrypto (commented):
See comments before approval

lib/firmware/ethereum.c
mcchadwick: That's right, tx has to have either a gas price set (non-EIP-1559) or max_fee_per_gas (EIP-1559.)

lib/firmware/ethereum.c
mcchadwick: I think that's a better idea. I'll make the change.

markrypto (approved):

PR #294, opened 2021-08-11
remove remaining MX code

mrnerdhair:
This gets rid of all the remaining message-processing code for Exchange-type messages, and drops support for decoding them.


lib/firmware/signing.c
```diff
@@ -953,15 +953,6 @@ static bool signing_check_output(TxOutputType *txoutput) {
     send_fsm_co_error_message(co);
     signing_abort();
     return false;
-  } else if (co < 0) {
```
reidrankin, line 956: these are dead code; `TXOUT_COMPILE_ERROR` is `0`, so the `co <= TXOUT_COMPILE_ERROR` branch above will always supersede these

unittests/firmware/coins.cpp
```diff
@@ -173,7 +173,7 @@ TEST(Coins, BIP32AccountName) {
                  {0x80000000 | 44, 0x80000000 | 931, 0x80000000 | 69, 0, 0},
                  5,
                  true,
-                 "THORChain Account #69\nAddress #0"}};
+                 "THORChain Account #69"}};
```
reidrankin, line 176: The `Address #n` was originally specified in a0250ff, and I don't think it was the originally intended behavior.

reidrankin:
Any thoughts on this? Appropriate for next release or not? If so, I'll merge the deps and re-request review.

markrypt0 (approved):

PR #295, opened 2021-08-18
Fix EIP-1559 transaction signing

mcchadwick:


mcchadwick:
With these changes, the KeepKey will produce decodable EIP-1559 transactions. The following tests still need to be performed:

- Ensure that the signed transactions will broadcast successfully
- Ensure that the transactions signed by KeepKey match those signed by hdwallet-native

markrypto (approved):

PR #299, opened 2021-09-10
Bugfix 1559 signatures

markrypto:


reidrankin (changes requested):
We should hard-fail if `ethereum_tx_type` isn't `ETHEREUM_TX_TYPE_LEGACY` or `ETHEREUM_TX_TYPE_EIP_1559`; otherwise looks good

mvpratt (approved):

lib/firmware/ethereum.c
```diff
@@ -651,11 +653,11 @@ void ethereum_signing_init(EthereumSignTx *msg, const HDNode *node,
 
   /* Ethereum tx type */
   if (msg->has_type) {
-    if (msg->type == 0 || msg->type == 1 || msg->type == 2) {
+    if (msg->type == ETHEREUM_TX_TYPE_EIP_1559) {
```
reidrankin, line 656: I nuked `ETHEREUM_TX_TYPE_LEGACY` too, because EIP-2718 makes it clear that a type of `0` is a _legal_ identifier for a new, non-legacy EIP-2718 transaction type, and simply not setting the field avoids this ambiguity.

markrypto: Looks like this change caused a bunch of test failures.

reidrankin: (working on that -- python-keepkey doesn't like my type=None solution)

reidrankin: sigh, it was just a stupid typo. have now updated python-keepkey to match, and tests should pass.

markrypto: Cool, thanks

reidrankin (approved):

lib/firmware/ethereum.c
reidrankin: Just realized this will break [hdwallet-keepkey](https://github.com/shapeshift/hdwallet/blob/bc7ad98972250855b213af076133b444d5c923e0/packages/hdwallet-keepkey/src/ethereum.ts#L54). Frankly, I don't think the type field is needed at all -- the presence/absence of maxFeePerGas allows us to infer that it's london/legacy, and denormalizing it just means more edge-cases where things can break.

What do you think about removing the field altogether?

markrypto: Crap. So revert this change and we'll deal with it in another PR. We need to get this out asap.

markrypto: Should be done with a separate PR with mcchadwick reviewing it.

PR #310, opened 2022-04-12
Add Cosmos staking/delegation support

pastaghost:


mrnerdhair (commented):
From a security perspective, this LGTM. That said, I can't give an approving review yet because I haven't been able to verify that the implementation is correct. To solve that we'll need to sort out the the amino broadcast problem so we can check that the signatures work on-chain.

mrnerdhair (approved):
We've [generated](https://github.com/shapeshift/hdwallet/pull/510/commits/468f20534145cf09c25c26d2caa8723dab742479) reference TXes with Keplr, confirmed they execute on-chain, and that KK generates matching signatures. There was a bug caught in that the amino type name had to be `MsgWithdrawDelegationReward` instead of `MsgWithdrawDelegatorReward`, and that the amount field in that message needed to be optional. This is G2G now.

PR #323, opened 2022-08-18
Symbol display

markrypt0:
Provides EIP-712 verification functionality.


lib/firmware/fsm_msg_ethereum.h
```diff
@@ -285,3 +285,34 @@ void fsm_msgEthereumSignTypedHash(const EthereumSignTypedHash *msg) {
...
+    fsm_sendFailure(FailureType_Failure_Other, _("Invalid EIP-712 types property string"));
+    return;
+  }
+
+  const HDNode *node = fsm_getDerivedNode(SECP256K1_NAME, msg->address_n,
+                                          msg->address_n_count, NULL);
+  if (!node) return;
+
+  uint8_t pubkeyhash[20] = {0};
+  if (!hdnode_get_ethereum_pubkeyhash(node, pubkeyhash)) {
```
pastaghost, line 306: If no pubkey hash is returned, the device will just return to the home screen here. Is this a failure mode for the function? If so, should we use a call to `fsm_sendFailure()` to notify the upstream application?

tools/bootloader/usb_flash.c
```diff
@@ -159,12 +159,11 @@ static bool should_restore(void) {
 }
 
 static bool isUpdateRequired(int signed_firmware) {
+  (void)signed_firmware;
```
pastaghost, line 162: What's going on here?

lib/board/resources.c
```diff
@@ -1830,6 +1830,56 @@ static const uint8_t unplug_data[395] = {
 static const Image unplug_image = {45, 27, sizeof(unplug_data), unplug_data};
 const AnimationFrame unplug_frame = {208, 21, 20, 100, &unplug_image};
 
+
+/* --- General image data, 20x40 ---------------------------------------------- */
+static const uint8_t ethereum_icon_data[367] =
```
pastaghost, line 1835: What do you think about standardizing the icon bitmap size and optionally passing signed icon data to the device dynamically? With the dynamic strategy, we'd store the bitmap data in globally-accessible context and keep it available for the duration of  the signing interaction. We could keep a few default icons like this one in flash to fall back on in general cases.

lib/firmware/fsm_msg_ethereum.h
markrypt0: The failure message is initiated in fsm_getDeriveNode().

tools/bootloader/usb_flash.c
markrypt0: An unused variable will generate a compiler warning. Compiler warnings are treated as errors in our build. (void) is the technique used to keep an unused var in the code but prevent the warning. This is some legacy stuff that someone thought might be useful at some point though we aren't currently doing anything with it.

pastaghost (approved):

PR #330, opened 2022-12-16
feat: add optimism native token

pastaghost:


BitHighlander:
old token. no one uses it.

PR #336, opened 2023-01-21
Allow unsigned rel

markrypt0:


markrypt0:
This PR does two things:  Allows unsigned release build firmware to run without checking for a recognized bootloader, and fixes the display when adding accounts with long auth secrets.

pastaghost:
> This PR does two things: Allows unsigned release build firmware to run without checking for a recognized bootloader, and fixes the display when adding accounts with long auth secrets.

The PR contents look sane to me; why do we want to allow running unsigned FW without the bootloader check?

markrypt0:
> The PR contents look sane to me; why do we want to allow running unsigned FW without the bootloader check?

There is a case I always run into, where I have a beta bootloader running and I can't run release build firmware for testing and this fixes it.

Really I'm looking for a reason NOT to allow this and I can't think of any. We already allow unsigned fw using signed bootloaders, any other bootloader is basically not keepkey. 

Signed firmware should continue to check for known good bootloaders in case there is some kind of bootloader downgrade attack.

pastaghost (approved):

pastaghost:
> 

Sounds good to me, senor. ✅

mrnerdhair:
There is a potential issue here: as we learned from [VULN-21020](https://blog.inhq.net/posts/keepkey-glitching-vuln-21020/), we should avoid re-running `signatures_ok()` in the firmware, because relatively simple glitch attacks can force a `SIG_FAIL` result even when the firmware is correctly signed. I don't immediately see an issue here that could be caused by this codepath being manipulated, but it's probably good practice to use a check like `*((uint8_t *)FLASH_META_SIGINDEX1) != 0` instead of `SIG_OK == signatures_ok()` here so we're sure that future design changes don't accidentally forget to consider such a case.

PR #371, opened 2026-03-11
fix(zcash): update branch ID from Sapling to NU6

BitHighlander:
## Summary
- Updates Zcash branch ID constant from Sapling (`0x76B809BB`) to NU6 (`0xC8E71055`) in `lib/firmware/signing.c`
- Single-line change, zero new dependencies

## Test plan
- [ ] Build firmware with Zcash signing enabled
- [ ] Verify Zcash transactions use NU6 branch ID
- [ ] Confirm no regression on other UTXO chains


BitHighlander:
Recreating with better description

PR #389, opened 2026-03-20
feat(policy): add blind signing policies, default disabled

BitHighlander:
## Summary
- Add `EthBlindSign` and `SolBlindSign` policies to the policy table, both defaulting to **disabled**
- Add hard gate in `ethereum.c` — arbitrary contract data signing is blocked unless `EthBlindSign` is enabled on-device
- Users must explicitly enable via `ApplyPolicies` which requires on-device button confirmation + PIN

## Test plan
- [ ] Flash firmware, verify blind signing ETH contract data is rejected by default
- [ ] Enable `EthBlindSign` via ApplyPolicies, confirm device prompts for button + PIN
- [ ] After enabling, verify ETH contract data signing works
- [ ] Verify `SolBlindSign` policy is present and defaults to disabled (enforcement comes with Solana merge)

### DIFF
--- a/include/keepkey/firmware/policy.h
+++ b/include/keepkey/firmware/policy.h
@@ -32,6 +32,8 @@ static const PolicyType policies[] = {
     {true, "Pin Caching", true, true},
     {true, "Experimental", true, false},
     {true, "AdvancedMode", true, false},
+    {true, "SolBlindSign", true, false},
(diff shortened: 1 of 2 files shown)


BitHighlander:
Accidental PR, closing — should target fork

PR #396, opened 2026-03-30
fix: BIP-39 per-word validation during cipher recovery

BitHighlander:
## Summary

During cipher-based recovery, the firmware accepted any character sequence without validating against the BIP-39 wordlist until the full mnemonic was assembled. A typo on word 8 of 24 wouldn't be caught until the end.

Now validates each word against the BIP-39 wordlist as it's completed, showing an OLED warning immediately if the word isn't valid.

## Changes

- `recovery_cipher.c`: Add per-word BIP-39 validation after each word completion
- `recovery_cipher.h`: Declare validation helper
- `bip39.c` / `bip39.h`: Add `bip39_word_is_valid()` lookup function
- `recovery_cipher.c`: Widen `prev_info` buffer 16→24 bytes for `-Wformat-truncation`

## Test plan

- [ ] CI builds pass (ARM + emulator)
- [ ] Cipher recovery with valid words proceeds normally
- [ ] Invalid word triggers on-device warning


BitHighlander:
Moving to fork — fixing python-keepkey test gating before reopening upstream

PR #404, opened 2026-03-31
feat: BIP-85 child mnemonic derivation (display only)

BitHighlander:
## Summary
- BIP-85 child mnemonic derivation — display only, never sent over USB
- Words shown on OLED one page at a time for user to write down
- 7 files, +272 lines

## Test plan
- [ ] CI green
- [ ] BIP-85 derivation test passes in emulator


BitHighlander:
Moving to fork — fixing python-keepkey test gating before reopening upstream

PR #415, opened 2026-04-03
fix(ci): pin clang-format to version 18 LTS

BitHighlander:
## Summary
Pin clang-format to version 18 (Ubuntu 24.04 LTS) instead of LLVM nightly.

The `llvm-toolchain-noble main` repo installs the latest nightly (currently v23), which produces different formatting than v18. This causes branches that format correctly locally to fail CI lint.

## Changes
- `ci.yml`: install `clang-format-18` from `llvm-toolchain-noble-18` instead of unpinned `clang-format` from nightly
- Symlink `/usr/bin/clang-format` to the pinned version

## Test Plan
- [ ] CI lint-format passes on develop
- [ ] Feature branches stop getting spurious lint failures

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,11 +52,12 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha || github.sha }}
 
-      - name: Install clang-format
+      - name: Install clang-format-18 (pinned LTS)
         run: |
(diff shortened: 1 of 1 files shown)


BitHighlander:
Closing — develop was formatted with v23 nightly, pinning to v18 breaks it. Will format feature branches to match instead.

PR #435, opened 2026-07-07
7.15 (2/5): New chain support — Hive (SLIP-0048), Zcash Orchard, Ripple memo, THORChain any-denom

BitHighlander:
**7.15 PR 2 of 5** (stacks on 1/5). Net-new chain families independent of clear-signing. Bumps trezor-firmware→`0ea97b09` for Orchard (Pallas/Sinsemilla/RedPallas/ZIP-316). Pins: device-protocol `up/release-protocol@33521a8`, python-keepkey `reconcile/upstream-sync@1674346` (keepkey upstream branches). Reconstituted + build-verified (kkemu + firmware-unit green); the 5-PR stack reproduces the 7.15 line byte-for-byte.


BitHighlander:
Superseded by #440. Relinearized onto `release/7.15.0-pr*` branches so the existing `release/**` CI trigger runs full CI on every PR in the stack (the old `7.15/*` bases weren't in the workflow's trigger filters, so only the bottom PR got CI). Same content, byte-for-byte.

PR #436, opened 2026-07-07
7.15 (3/5): Device robustness — BIP-85, BIP-39 recovery validation, fault-injection hardening

BitHighlander:
**7.15 PR 3 of 5** (stacks on 2/5). Independent robustness features. Pins: device-protocol `up/release-protocol@33521a8`, python-keepkey `reconcile/upstream-sync@1674346` (keepkey upstream branches). Reconstituted + build-verified (kkemu + firmware-unit green); the 5-PR stack reproduces the 7.15 line byte-for-byte.

### DIFF
--- a/include/keepkey/firmware/app_layout.h
+++ b/include/keepkey/firmware/app_layout.h
@@ -124,7 +124,8 @@ void layout_zcash_address_text_notification(const char* desc,
                                             const char* address,
                                             NotificationType type);
 void layout_pin(const char* str, char* pin);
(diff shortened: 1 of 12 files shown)


BitHighlander:
Superseded by #441. Relinearized onto `release/7.15.0-pr*` branches so the existing `release/**` CI trigger runs full CI on every PR in the stack (the old `7.15/*` bases weren't in the workflow's trigger filters, so only the bottom PR got CI). Same content, byte-for-byte.

PR #434, opened 2026-07-07
7.15 (1/5): EVM clear-signing core — signed-metadata, Insight, EIP-1559/RLP/EIP-712, drainer hardening

BitHighlander:
**7.15 upstream landing, PR 1 of 5 (stacked; merge bottom-up).** The signed-metadata trust system + the EVM hardening it rides on: v1/v2 static-schema on-device calldata decode, runtime-loaded signers (phase-1 warning), LoadClearsignSigner (+icon), Insight; EIP-1559 zero-priority RLP, RLP length-strip, EIP-712 cancel/strtoll validation, token chain-id; 0x/THORChain-router drainer hardening (owns thortx wholesale incl. Maya EVM display). 75/75 unit tests. Pins: device-protocol `up/release-protocol@33521a8`, python-keepkey `reconcile/upstream-sync@1674346` (keepkey upstream branches). Reconstituted + build-verified (kkemu + firmware-unit green); the 5-PR stack reproduces the 7.15 line byte-for-byte.

PR #439, opened 2026-07-08
7.15 (1/5): EVM clear-signing core + Ethereum signing hardening

BitHighlander:
**1/5 of the 7.15 stack — merge bottom-up, this one first.** Base: `develop`.

The signed-metadata trust system for EVM contract calls, plus the Ethereum signing-path hardening it rides on. Everything here is EVM; the other chains stack on top in later PRs.

## Signed-metadata clear-signing
- **Insight** — signed-metadata clear-signing with an AdvancedMode blind-sign gate: trusted metadata is signed; anything unsigned falls back to a blind-sign warning.
- **Runtime-loaded signers (phase 1)** — a signer catalog loadable at runtime, always behind a mandatory warning screen; dropped on device wipe.
- **v2 static schema** — the device decodes the calldata it signs from a static schema blob (no per-tx online signer) and renders the WHAT as `STRING` / `TOKEN_AMOUNT`. Decode state resets per `matches_tx` so a stale proof can't carry into the next tx. `LoadClearsignSigner.icon` gets a firmware-side nanopb `max_size`.

## Ethereum signing correctness (ported + hardened from alpha)
- **EIP-1559** — always hash the priority fee and reject tx-type / fee-field mismatches, so the stage-1 preview and stage-2 signature can't diverge.
- **RLP** — strip leading zeros in integer fields and in the list-length calc so the length prefix matches the bytes actually hashed (a mismatch here recovers the wrong signer).
- **EIP-712** — fix cancel propagation; validate `strtoll` input.
- **Drainer hardening** — pin the THORChain router to the current v4 address + validate memo offset; tighten 0x / clear-sign handler gating (ports alpha #255 / #260 / #261).
- Widen `TokenType.chain_id` to 32-bit; label Maya EVM txs "Maya data", not "Thorchain data".

## Verification & pins (whole stack)
75/75 firmware unit tests here; the stack builds green (kkemu + firmware-unit) at every step. Pins are identical across all five PRs — device-protocol [`keepkey/device-protocol@33521a8`](https://github.com/keepkey/device-protocol), python-keepkey [`keepkey/python-keepkey@e0587c06`](https://github.com/keepkey/python-keepkey) (the latter version-gates the Solana-v0 tests to 7.15). PR2 and PR5 additionally bump trezor-firmware (Orchard, then `AES_SMALL_TABLES`).

## Stack (merge order)
1. **this** — EVM clear-signing core → `develop`
2. new chains — Hive, Zcash Orchard, Ripple, THORChain any-denom
3. device robustness — BIP-85, BIP-39 recovery validation, fault-injection hardening
4. per-chain clear/blind-sign — TRON, Solana v0, TON, Maya / THORChain
5. build variants + version 7.15.0

The five branches are a clean linear relinearization of the 7.15 delta; merged in order they reproduce the 7.15 line exactly. Replaces #434.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "deps/device-protocol"]
 path = deps/device-protocol
 url = https://github.com/keepkey/device-protocol.git
-branch = master
+branch = up/release-protocol
 [submodule "deps/trezor-firmware"]
 path = deps/crypto/trezor-firmware
 url = https://github.com/keepkey/trezor-firmware.git
@@ -14,7 +14,7 @@ url = https://github.com/keepkey/code-signing-keys.git
 [submodule "deps/python-keepkey"]
 path = deps/python-keepkey
 url = https://github.com/keepkey/python-keepkey.git
-branch = master
+branch = reconcile/upstream-sync
 [submodule "deps/qrenc/QR-Code-generator"]
 path = deps/qrenc/QR-Code-generator
 url = https://github.com/keepkey/QR-Code-generator.git
--- a/include/keepkey/board/confirm_sm.h
+++ b/include/keepkey/board/confirm_sm.h
@@ -95,6 +95,10 @@ bool confirm(ButtonRequestType type, const char* request_title,
              const char* request_body, ...)
     __attribute__((format(printf, 3, 4)));
 
+bool confirm_with_icon(ButtonRequestType type, IconType iconNum,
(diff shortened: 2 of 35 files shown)


BitHighlander:
Closing — these were pushed from the fork-develop rehearsal branches and carry rehearsal/internal commit messages. Realigning on the release SOP before re-submitting clean upstream PRs.

PR #446, opened 2026-07-08
7.15 (3/5): Device robustness — BIP-85, BIP-39 recovery validation, fault-injection hardening

BitHighlander:
**3/5 of the 7.15 stack.** Stacks on 2/5 (`release/7.15.0-pr2-new-chains`).

Independent device-robustness work — no chain or clear-signing dependencies.

- **BIP-85** — child mnemonic derivation (display only).
- **BIP-39 recovery** — per-word validation during cipher recovery, so a mistyped word is caught on entry instead of producing a silently wrong seed.
- **Fault-injection hardening** — the signature-verification path is hardened against glitch / fault attacks.
- Plus the small build/lint fixes these need to pass ARM `-Werror` and cppcheck.

Same base pins as 1/5. Builds green.


BitHighlander:
Superseded by #450 — a single consolidated PR carrying the complete, current 7.15.0 content (this stack had drifted ~5.8k lines behind the release line). Review history for this slice remains here; the branch is left in place.

Issue #456, opened 2026-08-20
Osmosis: the screen names one asset and the signature commits to another


BitHighlander:
Reproduced on **signed v7.14.1** (`sha256 f40fe1b74949a9e091269536b27cd53bbb8a5c1c428a6c36ff7d5c6bf6845f46`, key slots 1/2/3).

## What happens

Send an Osmosis `MsgSend` of `123456789012345678 uion`:

| | |
|---|---|
| sent | `123456789012345678 uion` |
| **on screen** | `123456790519087104.000000` |
| **signed** | `"denom":"uosmo"` |

Two independent defects in one confirmation.

## 1. The amount is wrong past the 8th significant digit

    fsm_msg_osmosis.h:143   float amount = atof(msg->send.amount);
    fsm_msg_osmosis.h:151   snprintf(..., "%.6f %s", amount, denom);

`float` is 32-bit. `float32(123456789012345678)` is exactly `123456790519087104`,
which is what the OLED shows — the digits after the 8th are **wrong**, not
rounded. The `.000000` tail is on a denom that was never divided, because only
`uosmo` is decimalised.

## 2. The signed document names a different asset than the screen

    osmosis.c:78    "\",\"denom\":\"uosmo\"}]"
    osmosis.c:131   "\"amount\":[{\"amount\":\"%s\",\"denom\":\"uosmo\"}]"

`uosmo` is hardcoded into the amino document that gets signed, regardless of the
denom that was displayed. The screen is built from the host's denom; the
signature is built from a constant. They are not the same source, so they can
disagree — and for any non-`uosmo` denom they do.

The device signed.

## Fix

- decimalise only `uosmo`; print every other denom as the exact signed integer
  next to the exact denom, with no `.000000` tail
- build the amino document from the same denom that was displayed

Fixed in the 7.14.2 security line.

Issue #455, opened 2026-08-20
Authenticator: cancelling a wipe destroys the TOTP secrets anyway


BitHighlander:
Reproduced on **signed v7.14.1** (`sha256 f40fe1b74949a9e091269536b27cd53bbb8a5c1c428a6c36ff7d5c6bf6845f46`, key slots 1/2/3).

## What happens

    getAccount:0                 -> 'keepkey.com:alice'        account present
    wipeAuthdata + host Cancel   -> Success                    the wipe PROCEEDED
    getAccount:0                 -> Failure 'Account not found' secret is GONE

**The user's abort executed as a commit.** Cancelling the wipe is what destroyed
every authenticator secret on the device.

## Why it happens

`wipeAuthData()` calls `confirm()` and discards the result, then wipes
regardless. The same pattern affects the add and delete paths.

There is no reject button on this device, so a `false` from `confirm()` means
exactly one thing: the host sent a `Cancel` and the screen was taken away. That
is a refusal, and it was being ignored.

## Corroborating detail

The pre-fix warning text read:

> Do you want to PERMANENTLY delete all authenticator accounts?
> **If not, unplug Keepkey now.**

That second sentence only makes sense if cancelling does not work — which is
what the code did. It is a comment on the defect, written into the UI.

## Fix

Gate on the result:

```c
if (!confirm(ButtonRequestType_ButtonRequest_Other, "Confirm Wipe Authdata",
             "Do you want to PERMANENTLY delete all authenticator accounts?")) {
  return CANCELED;
}
```

and drop the "unplug Keepkey now" sentence, which is no longer true.

Verified after the fix with the same before/after read: the cancel returns
`Failure 'Action cancelled'` and `getAccount:0` still returns
`keepkey.com:alice`.

Fixed in the 7.14.2 security line.

Issue #454, opened 2026-08-20
THORChain: memo truncated by one character, and trailing fields signed without being shown


BitHighlander:
Reproduced on **signed v7.14.1** (`sha256 f40fe1b74949a9e091269536b27cd53bbb8a5c1c428a6c36ff7d5c6bf6845f46`, key slots 1/2/3).

Two separate defects on the same path. **They need different reproducers** —
testing one with the other's input produces a false negative, which happened
during verification here.

## 1. The last character of the memo is dropped

59-byte memo, via a BTC OP_RETURN swap:

    SWAP:ETH.ETH:0x41e5560054824ea6b0732e656e3ad64e20e94e45:420

**Screen reads `Confirm limit 42`.**

`strlcpy(memoBuf, swapStr, size)` copies `size - 1` bytes from a buffer the
OP_RETURN caller does not NUL-terminate, so the final character is lost. A
limit of 420 displayed as 42.

## 2. Trailing memo fields are signed and never displayed

65-byte memo — what a ShapeShift-style host actually sends:

    SWAP:ETH.ETH:0x41e5560054824ea6b0732e656e3ad64e20e94e45:420:ss:75

The device shows the asset, destination and limit. **No screen names the
affiliate or the 75 bps fee.** The signed transaction ends:

    ...3a3432303a73733a3735     = ":420:ss:75"

Those bytes are in the signature. The user is never shown the fee being taken.

Everything after the limit — affiliate name, affiliate fee in bps, aggregator
routing — was hashed with nothing on screen.

## Scope — stated precisely

The destination and amount **are** shown correctly. What is hidden is the
affiliate tail. This is an undisclosed fee, not an undisclosed recipient.

## Fix

- `memcpy` into a memzero'd buffer instead of `strlcpy` from a
  non-NUL-terminated source, so the last byte survives
- a `strtok` loop over the remaining `:`-separated fields, issuing one confirm
  per field with body `Additional memo field` / `<field>`

Verified after the fix: two extra screens naming `ss` and `75`, and the signed
bytes are byte-identical to the pre-fix fixture for a legacy memo — the parser
does not mutate what gets hashed.

Fixed in the 7.14.2 security line.

Issue #453, opened 2026-08-20
EVM: an omitted chain_id yields a pre-EIP-155 signature the user cannot see


BitHighlander:
Reproduced on **signed v7.14.1** — the current release — not on a development build.

    firmware.keepkey.bin
      sha256 (full signed file) f40fe1b74949a9e091269536b27cd53bbb8a5c1c428a6c36ff7d5c6bf6845f46
      sha256 (payload, tail -c +257) 73a880682d00c725cf9b38347ed0f08f4c2214c2c0f54559739f9d939f6275ae
      tag commit 5482e7366a36e81336074f53a9defb6cff45ed72, key slots 1/2/3

## What happens

Send `EthereumSignTx` with `chain_id` unset. The device signs, and returns:

    signature_v = 27

`v = 27/28` is a pre-EIP-155 signature. It carries no chain binding, so it is
valid on **every** EVM chain — the same transaction can be replayed anywhere the
account exists.

## What the user sees

Nothing that would let them catch it. With no `chain_id`,
`ethereumFormatAmount()` matches no case, so:

- the amount renders as a bare number with **no ticker**
- **no screen names a network**

Confirmed on the OLED during reproduction.

## Why it happens

The bounds check sat inside `if (msg->has_chain_id)`, so omission fell through
to `chain_id = 0` and execution reached both `confirm()` calls.

## Scope — stated precisely

This is **not** key disclosure and **not** a silent transfer. The user does
approve *a* transaction. What they cannot see is which chain it binds to, and
the answer is all of them.

## Fix

Hoist the check out of the `has_chain_id` guard so an absent field is treated as
`0` and refused before the first `confirm()`:

```c
chain_id = msg->has_chain_id ? msg->chain_id : 0;
if (chain_id < 1) {
  fsm_sendFailure(FailureType_Failure_SyntaxError, _("Chain Id out of bounds"));
  ethereum_signing_abort();
  return;
}
```

With that in place the refusal is the **first** reply on the wire and no
`ButtonRequest` is emitted at all — verified by reading the raw first response,
which proves no screen was drawn rather than inferring it from a photograph.

Fixed in the 7.14.2 security line.

PR #457, opened 2026-08-20
release: 7.14.2 security line — the device shows what it signs

BitHighlander:
Opens the 7.14.2 security line against `develop` (`1af2ffe7de`, = v7.14.1).

Every defect here is one shape: **the device hashed bytes it never rendered, or
rendered text it could not vouch for.**

## Issues closed

Filed separately with reproductions on **signed v7.14.1**, so each can be
verified against the firmware users actually run rather than a development tree:

- #453 — EVM: an omitted `chain_id` yields a pre-EIP-155 signature
- #454 — THORChain: memo truncated by one character, and trailing fields signed unseen
- #455 — Authenticator: cancelling a wipe destroys the TOTP secrets anyway
- #456 — Osmosis: the screen names one asset and the signature commits to another

Also in scope: EIP-712 (both entry points closed by default), blind-sign gates
for TON/TRON/ETH calldata, message-display disclosure, 0x calldata completeness,
Solana SPL amount binding, and the confirmation renderer's fit measurement.

## Evidence

**Reproduced on signed v7.14.1** before the fixes:
`f40fe1b74949a9e091269536b27cd53bbb8a5c1c428a6c36ff7d5c6bf6845f46`, tag commit
`5482e736`, key slots 1/2/3. Both the full-file and payload hashes were checked
against the release's own `HASHES.txt`, and the device reported the matching
`firmware_hash`.

**Verified on hardware after the fixes** — twelve test cards on 7.14.2 silicon.
Selected results:

- an injected `RecoveryDevice` mid-ceremony is refused before it can write
  anything; the `EntropyAck` that used to be consumed by the still-armed reset
  returns `Not in Reset mode`
- a cancelled passphrase confirmation is no longer cached, and the next request
  re-asks
- a cancelled authenticator wipe leaves the secret readable — proven with a
  before/after read across the operation
- an omitted `chain_id` is refused as the **first** reply with no `ButtonRequest`
  emitted, which proves the absence of a screen on the wire rather than
  inferring it from a photograph
- both EIP-712 endpoints refuse in **21 ms**; the structured disable is hardcoded,
  so enabling AdvancedMode does not re-open the withdrawn parser
- 0x clear-signing is bound to complete calldata and to both tokens resolving —
  the same calldata with one 32-byte word changed signs in one case and refuses
  in the other

**Rendering is now captured automatically.** The suites this release changes had
no OLED capture in CI: the emulator harness read the firmware version with
`grep -oP`, which BusyBox lacks, so it silently fell back to `7.14.0` and every
test gated to 7.14.1+ was excluded from the screenshot filter. Fixed, plus a
per-test audit that fails the build when a test declares screens it does not
produce. Captures went 345 → 381.

## Behaviour changes a host will notice

- **`EthereumSignTx` without `chain_id` is refused.** Hosts that relied on the
  implicit `0` must send an explicit chain id.
- **Structured EIP-712 (`Ethereum712TypesValues`) is disabled.** The legacy JSON
  parser could not guarantee that every displayed value was the canonical value
  being hashed, and one screen took its title from the attacker-supplied domain
  name. Withdrawn rather than shipped with a screen it could not vouch for.
- **`EthereumSignTypedHash` now requires AdvancedMode**, and shows an explicit
  blind-sign screen first.
- **A body too long for the screen is paged**, with `n/m` in the title. Reading a
  page costs a short click; only the final page takes a full hold.
- **Message signing renders non-printable bytes as escapes.** `Sign in to
  example.com` now displays as `Sign\x20in\x20to\x20example.com`. This is
  deliberate: `%s` stopped at the first NUL, so bytes after one were signed and
  never drawn.

## Submodules

- `deps/device-protocol` → `f2c3c005`, **identical to its upstream master**
- `deps/python-keepkey` → `bdb4444`, which is **44 commits ahead of pyk master
  and not yet merged**. It is the head of keepkey/python-keepkey#214, the single
  open PR carrying this release's test changes. That PR merges once this one is
  green — a reviewer cloning pyk master will not resolve this pin until then.

## What is NOT claimed

Stated because the review is worth more than the release:

- **Independent re-exploitation and a final adversarial round have not been
  done.** They are not self-certifiable; this review is how they happen.
- **The fit measurement covers `layout_standard_notification` only.**
  Custom-layout address and xpub bodies still clip with no warning. Known,
  documented, not fixed here.
- **One finding was filed and retracted** during this work — it rested on an
  inferred button press that had actually occurred. The refuting control took a
  minute and belonged before the issue was filed. Recorded in the release notes
  rather than quietly dropped.
- Two further issues in the tracker overstate their own reproductions and are
  deliberately **not** filed upstream until reworded to what was measured.

Full evidence: `docs/release/rc30-evidence/` (13 documents),
`docs/release/signed-repro/` (reproductions on signed firmware),
`docs/release/7.14.2.md` (gate status, measured rather than asserted).


BitHighlander:
Closing: wrong shape. This should be a branch in this repository into develop, not a PR from a fork head.

Reopening from an upstream branch shortly. No content changes — same tree, same evidence.

PR #462, opened 2026-08-23
release: 7.15

BitHighlander:
Stages **7.15** onto `develop`.

**Merge order matters: this sits on top of 7.14.3 (#461).** `develop` here is
7.14.1; #461 brings 7.14.2 + 7.14.3, and this branch was built on that base. Land
#461 first.

Version `7.15.0`, `STORAGE_VERSION` 17.

## Scope

Merges the last 7.15 commit on `alpha` (`0e494d104`). Alpha has since moved to
7.16, so **7.16 is deliberately excluded** — passkeys, CTAP2 and storage V20
(with 18/19 burned) are 45 files / +5,477 sitting cleanly on top of that
boundary. `STORAGE_VERSION` remaining 17 is the check that none of it leaked in.

## The merge

16 files conflicted, 43 hunks, resolved per hunk. Neither parent is a subset of
the other — merge-base `bd3a1d6e9`, and both sides did independent bitcoin-only
work — so this is a union nearly everywhere:

| side | brought |
|---|---|
| develop | taproot, the `KK_BITCOIN_ONLY` variant, the storage version band, the shipped-version floor, the `btc_only_locked` guard |
| 7.15 | zcash, clear-signing, hive, solana, release infrastructure |

Audited afterwards: every storage control present and coherent, `KK_BITCOIN_ONLY`
gating intact, taproot intact, no 7.15 identifier dropped, no leftover markers.

## Findings from independent review, all addressed

**[P1] Orchard signing bypassed the RNG health gate.** `ZCASH_PRIVACY=1` makes
Orchard reachable in the regular image, and `redpallas_sign_with_rsk()` draws its
spend-authorization nonce from plain, unchecked `random_buffer()`. **A repeated
nonce discloses the spend authorization key.** Describing this as inherited was
wrong: the crypto lived in the pinned dependency, but base `develop` never
compiled or exposed Orchard, so this merge is what makes the path reachable.

The draw is in vendored crypto and cannot be routed through
`random_buffer_checked()` from the firmware, so the gate is at the call site and
fails **closed** before every spend — the latched boot verdict plus a live
checked draw folded into the continuous SP 800-90B state immediately before the
nonce is taken. Probe bytes are discarded and zeroed; they are a health
observation, not key material.

This is **not** equivalent to the nonce itself being checked, and the code says
so. The complete fix is a checked draw inside RedPallas or an independently
reviewed deterministic nonce — both require changing the vendored crypto, and
should be done before Orchard is considered production-ready.

**[P2] The verified gitleaks binary was replaced by an unverified one.** A clean
merge unioned both parents: 8.30.1 was downloaded and checksummed, then 8.30.0
was fetched via `curl | tar` with no checksum, over the top, before the scan ran
— so the checksum proved nothing about the executable that ran. Now one version,
one archive, one verified digest, pinned at 8.30.0 (8.30.1 false-positives on
published BIP32 test vectors, and since every build job declares
`needs: secret-scan`, that **skips** the whole graph rather than failing it).

**[P2] The generated release checklist omitted signature verification.**
`docs/Release.md` requires `verify-signatures.py` before publishing and states
that `hash-manifest.sh --require-signed` is structural only — a signature region
holding one non-zero byte passes it. The checklist ran only the structural check.
The real 3-of-5 ECDSA verification is now required for every variant, before
upload.

## Submodule pins

| | pin | |
|---|---|---|
| `deps/device-protocol` | `8545cd5b` | what 7.15 builds against |
| `deps/python-keepkey` | `44d82efe` | python-keepkey#197 head |

The device-protocol pin is deliberately **not** device-protocol#112's head: #112
adds thirteen structured EIP-712 fields this tree has no `.options` entries for,
so nanopb emits `pb_callback_t` and every build fails. Verified by building.

## Verification, and what is NOT established

Both variants build clean: full 562,748 B, bitcoin-only 336,184 B.

**The taproot and storage-band work carried up from 7.14.3 IS verified.** On a
real K1-14AM: the band round-trips a power cycle (seed stamped `10017` → replug
→ same addresses), BIP-86 matches the **spec** vector, and a taproot spend signs
**byte-exact** against an independently computed BIP-340 vector. On the emulator,
both transition directions check out — bitcoin-only → multi-chain is REFUSED
with the flash image **byte-identical** afterwards, and multi-chain →
bitcoin-only is allowed with the ordinary "already initialized" refusal rather
than the band lock.

**Orchard has NO hardware verification, and that is the gap that matters here.**
The RNG gate added in this PR fails closed on a sick source, but it has never
run on a device, and the nonce itself still comes from unchecked
`random_buffer()` inside vendored crypto. Orchard should not be treated as
production-ready until the nonce is drawn through a checked interface — or an
independently reviewed deterministic construction — and that path is exercised
on hardware.

**Also still unproven:** that storage survives a real firmware swap. That is the
bootloader's preserve path, and it cannot be shown with unsigned images — the
bootloader wipes on any unsigned transition by design
(`docs/StorageVersionGate.md`). It needs a signed build on a production device.


lib/firmware/fsm_msg_zcash.h
```diff
@@ -0,0 +1,1661 @@
...
+    0xf4, 0xbe, 0xd7, 0x43, 0x91, 0xee, 0x0b, 0x5a, 0x69, 0x94, 0x5e,
+    0x4c, 0xed, 0x8c, 0xa8, 0xa0, 0x95, 0x20, 0x6f, 0x00, 0xae};
+
+/* ZIP-244 empty-bundle digest: BLAKE2b-256 of the empty string personalized
+ * "ZTxIdOrchardHash". A v6 (Ironwood) transaction streams and verifies only its
+ * Ironwood actions, so its Orchard bundle MUST be empty -- otherwise the device
+ * would sign a sighash committing to an Orchard bundle it never inspected.
+ * Sapling and the transparent bundle are already pinned this way above; Orchard
+ * was the one unused pool taken on trust from the host. */
+static const uint8_t EMPTY_ORCHARD_DIGEST[32] = {
```
BitHighlander: [P1] Use the v6 empty-Orchard digest here. ZIP-229 changes the v6 Orchard component personalization to `ZTxIdOrchardH_v6`, including for an empty bundle. This literal is the v5 hash `BLAKE2b-256("ZTxIdOrchardHash", "")` (`9fbe...e2c1`); v6 requires `a3367d2fdea2910159fc5026e9bf1fccd3e28ce5e6de46bfb71587230eea9515`. The Ironwood gate therefore rejects a correct v6 PCZT, while any request satisfying this gate feeds the wrong child into the v6 sighash and produces signatures the network cannot validate. Please derive this constant with the ZIP-229 personalization and update the host fixture. Spec: https://github.com/zcash/zips/blob/main/zips/zip-0229.md#anchor-commitment-version-6

BitHighlander (commented):
Security review of the current 7.15 head. These comments are limited to reproduced code/spec mismatches and signed-display behavior.

lib/firmware/fsm_msg_zcash.h
```diff
@@ -0,0 +1,1661 @@
...
+  }
+
+  return true;
+}
+
+static bool zcash_compute_active_sighash(const uint8_t transparent_digest[32],
+                                         uint8_t sighash[32]) {
+  if (zcash_signing.transaction_v6) {
+    return zcash_compute_v6_shielded_sighash(
+        zcash_signing.header_digest, transparent_digest, EMPTY_SAPLING_DIGEST,
```
BitHighlander, line 477: [P1] The v6 sighash still receives the v5 empty-Sapling digest. ZIP-229 changes `sapling_digest_v6` to personalization `ZTxIdSaplingH_v6`; the empty digest is `5a94e6cb4afbb0aa43f70a20273a76d69a12acf5a78518b1a3f3236d09ad6fe1`, whereas `EMPTY_SAPLING_DIGEST` is the v5 `6f2f...00ae` value. Because this signer intentionally supports no Sapling bundle, it must select the canonical v6 empty digest here. As written, every v6 signature hash is wrong regardless of which shielded pool is active.

lib/firmware/fsm_msg_zcash.h
```diff
@@ -0,0 +1,1661 @@
...
+      blake2b_Final(&zcash_signing.memos_ctx, memos_hash, 32);
+      blake2b_Final(&zcash_signing.noncompact_ctx, noncompact_hash, 32);
+
+      /* V5 Orchard commits the anchor in the txid component. Transaction-v6
+       * Ironwood moves the anchor to the authorizing-data digest (ZIP-229),
+       * so the device deliberately omits it here. */
+      BLAKE2B_CTX orchard_ctx;
+      blake2b_InitPersonal(
+          &orchard_ctx, 32,
+          zcash_signing.is_ironwood ? "ZTxIdIronwd_H_v6" : "ZTxIdOrchardHash",
```
BitHighlander: [P1] A non-Ironwood `tx_version == 6` is accepted, but this finalizer still uses the v5 `ZTxIdOrchardHash` personalization, and `ironwood_component_digest` remains all zero because it is copied only when `is_ironwood`. ZIP-229 requires `ZTxIdOrchardH_v6` for the active Orchard component and the canonical empty Ironwood digest (`b9cfe643ce45b28c33190f0d5223e475972f2a149dc54404fd8365521f8416c5`) as the final aggregate child. Consequently a valid v6 Orchard request either fails the local component check or yields a network-invalid sighash. Please implement both v6 children, or explicitly reject v6 unless the supported Ironwood path is selected.

lib/firmware/fsm_msg_thorchain.h
```diff
@@ -217,10 +264,21 @@ void fsm_msgThorchainMsgAck(const ThorchainMsgAck* msg) {
   if (sign_tx->has_memo && !msg->deposit.has_memo) {
     // See if we can parse the tx memo. This memo ignored if deposit msg has
     // memo
-    if (!thorchain_parseConfirmMemo(sign_tx->memo, sizeof(sign_tx->memo))) {
-      // Memo not recognizable, ask to confirm it
-      if (!confirm(ButtonRequestType_ButtonRequest_ConfirmMemo, _("Memo"), "%s",
-                   sign_tx->memo)) {
+    /* strnlen, not sizeof -- see the deposit path above. */
+    ThorchainMemoResult memo_result = thorchain_parseConfirmMemo(
```
BitHighlander, line 275: [P1] Do not let the current ack payload suppress review of the transaction-level memo. `ThorchainMsgAck` is proto2 with independent `send` and `deposit` fields, and the validation/processing logic chooses `send` first. A host can therefore send a valid `send` plus an otherwise unused `deposit.has_memo = true`; this condition skips `sign_tx->memo` even though that memo was already hashed into the signed Amino JSON and can control routing for a vault transfer. Always review `sign_tx->memo` when present, and reject acks unless exactly one of `has_send` / `has_deposit` is set.

lib/firmware/fsm_msg_mayachain.h
```diff
@@ -218,17 +279,15 @@ void fsm_msgMayachainMsgAck(const MayachainMsgAck* msg) {
...
-                   sign_tx->memo)) {
-        mayachain_signAbort();
-        fsm_sendFailure(FailureType_Failure_ActionCancelled, NULL);
-        layoutHome();
-        return;
-      }
+    // Ignored if the deposit msg has a memo. Page the full raw memo as the
+    // sole gate -- strnlen, not sizeof; see the deposit path above for why
+    // there is no structured pass.
+    size_t memo_len = strnlen(sign_tx->memo, sizeof(sign_tx->memo));
```
BitHighlander, line 292: [P1] This has the same signed-memo disclosure bypass as the THORChain handler. The final ack can contain a valid `send` together with an unused `deposit.has_memo = true` because these are independent proto2 fields and the handler takes the send branch first. The enclosing condition then hides the transaction-level memo even though `mayachain_signTxInit` already committed it to the signed JSON. Review the top-level memo independently of the ack subtype, and reject acks that set both or neither subtype.

lib/firmware/fsm_msg_solana.h
```diff
@@ -407,6 +424,81 @@ void fsm_msgSolanaSignTx(const SolanaSignTx* msg) {
       return;
     }
 
+    /* KKSOLSW1: a provider may attest the accounts this message resolves
```
BitHighlander, line 562: [P2] The advertised `KKSOLSC1` path is never consumed by this production handler. `SolanaSignTx` exposes `schema_payload`, `schema_signature`, and `schema_signer_key_id`; the parser, matcher, attestor, and isolated unit tests exist; but `fsm_msgSolanaSignTx` references none of those fields and only handles the `KKSOLSW1` LUT annotation before showing the blind-sign warning. Thus every valid reusable schema transaction remains opaque/blind despite the protocol and release documentation. Please verify and parse the schema here, call `solana_schemaApplies`, and render the matched fields, or remove the shipping contract/docs until that path is wired.

lib/firmware/fsm_msg_zcash.h
BitHighlander: Correction after checking the live PR head: this is already addressed directly by a3da8284565c56f203e6f9a54d888fc2373ebb41, an ancestor of exact head effdcd9e9c16d9b7d3b80cd9d33ef1f602a610bc. Current v6 finalization selects ZTxIdOrchardH_v6 and binds EMPTY_IRONWOOD_DIGEST_V6 b9cfe643ce45b28c33190f0d5223e475972f2a149dc54404fd8365521f8416c5 for non-Ironwood v6. Resolving.

lib/firmware/fsm_msg_mayachain.h
BitHighlander: Correction after checking the live PR head: a3da8284565c56f203e6f9a54d888fc2373ebb41 is already in this PR. Current MayachainMsgAck validation rejects both/neither subtype via has_send == has_deposit, and the transaction-level memo is reviewed independently. Resolving.

lib/firmware/fsm_msg_solana.h
BitHighlander: Correction after checking the live PR head: the production KKSOLSC1 path is already wired directly in this PR by a3da8284565c56f203e6f9a54d888fc2373ebb41. fsm_msgSolanaSignTx calls solana_confirmAttestedSchema for opaque transactions; that helper verifies schema_payload, schema_signature, and schema_signer_key_id, parses the schema, calls solana_schemaApplies, and renders program, discriminator, arguments, and accounts before the blind-sign warning. The fork cb710 path is a later 7.16 certified model, not the missing fix for this 7.15 thread. Resolving.

lib/firmware/fsm_msg_zcash.h
BitHighlander: Correction after checking the live PR head rather than the fork port: this is already addressed directly in this PR by a3da8284565c56f203e6f9a54d888fc2373ebb41, which is an ancestor of exact head effdcd9e9c16d9b7d3b80cd9d33ef1f602a610bc. The current code defines EMPTY_ORCHARD_DIGEST_V6 as a3367d2fdea2910159fc5026e9bf1fccd3e28ce5e6de46bfb71587230eea9515 and selects ZTxIdOrchardH_v6. Resolving.

lib/firmware/fsm_msg_zcash.h
BitHighlander: Retraction confirmed against the live exact head and current ZIP-229: the empty top-level Sapling component remains canonical with ZTxIdSaplingHash. No code change is required. Resolving the retracted finding.

lib/firmware/fsm_msg_thorchain.h
BitHighlander: Correction after checking the live PR head: a3da8284565c56f203e6f9a54d888fc2373ebb41 is already in this PR. Current ThorchainMsgAck validation rejects both/neither subtype via has_send == has_deposit, and the transaction-level memo is reviewed independently. Resolving.

BitHighlander:
Correction and final live-head disposition, superseding the earlier version of this comment:

- PR #462 is open, mergeable, and green at exact head effdcd9e9c16d9b7d3b80cd9d33ef1f602a610bc.
- All five implementation findings were addressed directly in this PR by a3da8284565c56f203e6f9a54d888fc2373ebb41, which is an ancestor of the exact head. Live blame confirms the v6 Orchard/Ironwood digests, THOR/MAYA subtype and memo checks, and the production Solana KKSOLSC1 review path all come from that commit.
- The Solana production path is solana_confirmAttestedSchema, called from fsm_msgSolanaSignTx for opaque transactions. It verifies the schema attestation, parses and matches it, and renders program, discriminator, arguments, and accounts before the mandatory blind-sign warning. The fork cb710 path is a later 7.16 certified model and is not required to close this 7.15 finding.
- The sixth Sapling finding is retracted: current ZIP-229 keeps the empty top-level Sapling component under ZTxIdSaplingHash.
- The six review-thread replies have been corrected against the live head and are being resolved.

Fork staging is separately complete at develop da075b8cb717b56dc1023edb52c2ccdf171b8a08, with #604 merged, #613-#618 closed, zero open fork PRs, and full/Bitcoin-only aggregate gates green. That later 7.16 line should not be force-pushed wholesale into #462.

Review disposition: the six findings raised on #462 are addressed or retracted at exact head effdcd9e9.... From this review, no finding remains as an upstream merge blocker. This does not authorize signing or release.

PR #472, opened 2026-08-27
docs(ci): add Copilot code review instructions; fix secret-scan

pastaghost:
## Summary

Adds `.github/copilot-instructions.md`, which governs how Copilot conducts code reviews on this repository — process instructions, not code-quality rules. Along the way this fixes two pre-existing `secret-scan` failures that were breaking CI on `develop` and every PR branch.

### Copilot review instructions

- **Report everything in one pass** — review the complete diff, don't sample, and comment on every instance of a repeated problem rather than just the first.
- **Include minor and uncertain findings** — report low-severity and less-confident findings with the uncertainty stated plainly, and label each finding's severity.
- **Converge in as few rounds as possible** — one round ideal, five the ceiling.
- **Don't trade thoroughness for brevity** — more findings per review, not faster reviews.

### CI fix 1 — gitleaks false positive on submodule commit pins

`secret-scan` was already failing on `develop` before this branch. gitleaks flagged the submodule commits recorded in release/security docs:

```
- python-keepkey: `c406a1ba9120da410c356dbff7f4d4bd1e1758fa`;
```

A 40-char lowercase hex git SHA carries enough entropy to trip `generic-api-key`. Since the job checks out with `fetch-depth: 0` and scans all reachable history, this failed every branch, not just the one holding the doc.

Fixed with a scoped allowlist in a new `.gitleaks.toml` rather than a `.gitleaksignore` entry. `.gitleaksignore` keys on commit SHA, and **that doc line already exists under three different commits** (`6f3becc6`, `9b9faac3`, `558fa241`) across branches — so every rebase, cherry-pick, or merge reintroduces it under a new fingerprint. The existing `trezordevkey.pem` entry shows the same pattern.

`[extend] useDefault = true` keeps every stock rule active, and `condition = "AND"` scopes the allowlist to SHA-shaped strings inside markdown under `docs/` only. Verified:

| Case | Expected | Result |
|---|---|---|
| 40-hex SHA in `docs/*.md` | exempt | not flagged |
| Same SHA line outside `docs/` | still flagged | flagged |
| Non-SHA secret inside `docs/*.md` | still flagged | flagged |
| AWS / GitHub PAT / Slack tokens in `docs/` | still flagged | all flagged |

### CI fix 2 — unpinned gitleaks download returning HTTP 403

The install step resolved the version via `api.github.com/.../releases/latest`. Unauthenticated api.github.com requests are rate limited per runner IP, and the call began returning `curl: (22) 403`, failing the job before the scan ran.

Pinned to `8.30.1` (the version CI was already resolving to) and downloading the release asset directly — no API call. The tarball is verified against its published SHA-256 before unpacking, curl retries transient errors, and the step runs under `set -euo pipefail`. Pinning also makes results reproducible: a new gitleaks release can change the default rule set, which under `latest` would land in CI on an unrelated PR with no commit explaining it.

## Submodule bumps

- `deps/device-protocol` → `f2c3c005` — auto-request Copilot review on PR open/push
- `deps/python-keepkey` → `598097e1` — same CI workflow, plus the `release/7.14.2-tests` merge

## Test plan

Documentation and CI configuration only; no firmware code paths are touched.

`secret-scan` now passes on this branch, and the scan genuinely ran rather than passing vacuously:

```
gitleaks_8.30.1_linux_x64.tar.gz: OK
8.30.1
4471 commits scanned.
no leaks found
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,47 @@
+# Copilot code review instructions
+
+How reviews on this repository should be conducted. These govern the review
+process itself, not what counts as good code.
+
+## Report everything in one pass
+
+Report every finding you have in a single review. Do not hold findings back for a
+later round.
+
+- Review the complete diff — every changed file, every changed hunk. Do not stop
+  partway through.
+- Finding several problems early is not a reason to end the review. It is a reason
+  to keep going.
+- Do not sample. If a file has twelve problems, report twelve.
+- Comment on every instance of a repeated problem, not just the first. If the same
+  mistake appears in six places, mark all six — the author fixes what is marked, so
+  "and similar issues elsewhere" leaves five defects in the branch.
+
+## Include minor and uncertain findings
+
+Do not filter down to only the findings you are most confident about.
+
+- Report low-severity and minor issues alongside significant ones.
+- Report a finding you are less than certain about, and say plainly what you are
+  unsure of. A finding the author dismisses in ten seconds costs far less than an
+  entire extra review cycle.
+- Label each finding's severity so the author can triage quickly instead of having
+  to weigh every comment equally.
+
+## Converge in as few rounds as possible
+
+Every additional review round costs the author a full cycle. Target:
+
+- **One round** — ideal. Everything surfaced on the first review.
+- **Two or three rounds** — acceptable.
+- **Five rounds** — the absolute ceiling.
+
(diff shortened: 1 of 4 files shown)


BitHighlander (approved):
:shipit:

